PR #29, opened 2022-10-12
Simplified Chinese Translation

Sdarfeesh:


adil192:
Thanks for the help :)

PR #32, opened 2022-10-16
english translation: improvements for consistency

mamasch19:
I changed some of the english translations to be more in line with what they mean/what they do.
Most of it are just proposals, if you don't like them it's also fine with me.

- capital letters
- Ctrl + Key instead of Ctrl Key
- "All notes" as the title of for the browse section, similar to how "Recent notes" already is

### DIFF
--- a/lib/i18n/strings.g.dart
+++ b/lib/i18n/strings.g.dart
@@ -3,7 +3,7 @@
 /// Locales: 2
 /// Strings: 119 (59 per locale)
 ///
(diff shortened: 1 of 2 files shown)


adil192:
Thanks for your interest.
I tend to prefer sentence case (where only the first letter is capitalised) over title case in most scenarios, and I think the "Ctrl Key" format looks nicer though this is definitely subjective.
I do like "All notes" over "Browse" for consistency with "Recent notes" however, so I'll merge this and the "Confirm" bit if you're happy with that

PR #54, opened 2022-11-07
Update Simplified Chinese Translation

Sdarfeesh:


adil192:
Thanks, not sure why the build is failing but it's just at the signing stage so it's unrelated

PR #56, opened 2022-11-09
add Traditional Chinese

jhihyulin:


adil192:
Thank you!

adil192:
@jhihyulin Would you be interested in being @'ed when there's new strings that need to be translated? https://github.com/adil192/saber/issues/19#issuecomment-1306096416

jhihyulin:
> @jhihyulin Would you be interested in being @'ed when there's new strings that need to be translated? https://github.com/adil192/saber/issues/19#issuecomment-1306096416

No problem

Issue #64, opened 2022-11-13
When finger drawing is disabled, the last stylus stroke is removed when pinch zooming


takoyakituesday:
I've been using the app on a Galaxy Tab S6 Lite, and the last stylus stroke gets removed when pinch zooming, but only if finger drawing is disabled.

I think it's treating the last stylus stroke as the mark made from the finger during pinch zoom. If I'm understanding correctly, I think it can be fixed by just checking if finger drawing is enabled before removing the last stroke.

I tested it and it seems to work for me, but I'm not familiar enough with the app to know if there might be unintended side effects. I'll open a small PR. Feel free to reject if I'm missing something obvious.

Thanks for your work on this app, it's really nice to have a FOSS notes app with proper stylus support.

adil192:
Good spot and thanks for the help :)

PR #65, opened 2022-11-13
editor: do not remove stylus strokes on zoom gesture

takoyakituesday:
This seems to fix the issue I'm having here: #64 

Thanks for reviewing.


adil192:
LGTM!

Issue #78, opened 2022-11-16
Unable to set Traditional Chinese


jhihyulin:
The system language of my device is Traditional Chinese, but Simplified Chinese is displayed in the `v0.3.16`
<img src="https://user-images.githubusercontent.com/74854953/202306823-beb1645c-7340-4802-9967-dfa44bc747fb.jpg" alt="Screenshot_2022-11-17-06-13-41-39_fd8abdc593c9d125bcab274fe725aa2b" width=300>

adil192:
Please can you try the apk from here https://github.com/adil192/saber/actions/runs/3483700101 when it's finished building

jhihyulin:
> Please can you try the apk from here https://github.com/adil192/saber/actions/runs/3483700101 when it's finished building

The problem remains, the display is still Simplified Chinese.

adil192:
How about [Build 21](https://github.com/adil192/saber/actions/runs/3486238543)?

jhihyulin:
> How about [Build 21](https://github.com/adil192/saber/actions/runs/3486238543)?

No change.

adil192:
Hmmm, you might need to take this up with [slang](https://github.com/Tienisto/slang/issues)

Sdarfeesh:
> @Sdarfeesh please can you verify that Simplified Chinese still works with the apk from [Build 23](https://github.com/adil192/saber/actions/runs/3490019840)?

Simplified Chinese works fine

PR #79, opened 2022-11-17
 fix zh-TW display

jhihyulin:
#78
change to zh_Hant-TW zh_Hant-CN
Local test is normal


jhihyulin:
Local test

Screenshot_2022-11-18-00-00-40-75_fd8abdc593c9d125bcab274fe725aa2b

PR #87, opened 2022-11-26
German Translation

mamasch19:
Sorry it took until now, didn't find time before. Everything should be fine, checked all the strings again.


adil192:
Thank you and no need to apologise!

PR #128, opened 2022-12-10
Reduce risk for memory leakage

yurtemre7:
Hello,

when I started using your App I enjoyed the good and sleak performance, but over time, the App, specially the Editor started to kinda "lag" as if my strokes are in slow-mo? I thought immediatly of some memory leakage. The rule I learned as well is always dispose your controllers. My PR consists on it, I did test it, everything is still working.

And hey, I really like your project, feel free to contact me [emredev](https://t.me/emredev), I'd love to work with you on it.

See you,
Emre Yurtseven


adil192:
LGTM thanks!

PR #138, opened 2022-12-11
updates: fix deprecated function error and fix infinite loading after error on login for nc

yurtemre7:
This function is deprecated as stated in their implementation of [shareFiles](https://github.com/fluttercommunity/plus_plugins/blob/21f38017fefcb2ceddac275cd4d76e26aaaab9a2/packages/share_plus/share_plus/lib/share_plus.dart#L72).

Could have technical reasons or even security because they want to work with XFile, not File anymore.

Edit: The second commit is about an error I have discovered when an error occured, that the spinning of the login button wouldn't stop, setState was missing.

### DIFF
--- a/lib/components/nextcloud/login_group.dart
+++ b/lib/components/nextcloud/login_group.dart
@@ -81,7 +81,9 @@ class _LoginInputGroupState extends State<LoginInputGroup> {
     }
 
     try {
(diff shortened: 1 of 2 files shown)


adil192:
Nice finds!

PR #176, opened 2022-12-27
i18n: Update Russian

ShevT:


adil192:
Just letting you know you could edit `_missing_translations.json` instead of directly editing `strings_ru.i18n.json` if that's easier for you. You don't have to of course!

PR #240, opened 2023-01-16
i18n: Update Russian

ShevT:
P.S. Please return tabulations in translation files as it is done in strings.i18n.json. Now it has become much more difficult to compare strings.

### DIFF
--- a/lib/i18n/_missing_translations.json
+++ b/lib/i18n/_missing_translations.json
@@ -58,23 +58,7 @@
     }
   },
   "it": {},
-  "ru": {
-    "settings": {
-      "prefLabels": {
-        "locale": "App language"
-      },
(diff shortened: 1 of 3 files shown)


adil192:
The indentations of the json files is done automatically by `slang`. I've reindented `strings.i18n.json` to match that

adil192:
> Now it has become much more difficult to compare strings.

I'm not sure what you mean by this. Are you manually editing `lib/i18n/strings_ru.i18n.json` each time? 
If so, it would be easier to edit [`_missing_translations.json`](https://github.com/adil192/saber/blob/main/lib/i18n/_missing_translations.json) instead. See https://github.com/adil192/saber#extending-existing-languages

PR #276, opened 2023-01-28
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Hej, @adil192, I've done some minor fixes in Czech translation, too. I hope it will help with better UX (primarily user understanding).

There is one thing I got today. I have 5 GB of space on my Nextcloud account registered on your server without sending you any pense. So, I think it is a bug in descriptions (where standard storage has 500 MB in size) or somewhere else. Thanks for looking at it!

adil192:
Yeah I was midway through updating the baseline storage size and then got distracted and forgot to update GitHub :p

PR #292, opened 2023-02-01
i18n: Add Persian

mehdijahann:
Persian language ;)


adil192:
Hi, would you be interested in being mentioned when there's new strings that need to be translated? https://github.com/adil192/saber/issues/19#issuecomment-1306096416

PR #306, opened 2023-02-04
Remove web

adil192:


pauljako:
Would you elaborate, why you removed web support? It would be really useful for me :/

PR #310, opened 2023-02-04
Add Czech translation of background on main screen

ceskyDJ:


ceskyDJ:
@adil192, what does it mean/do? I tried it on the web version https://adil192.github.io/saber/#/home/recent and I wasn't able to find it out.

adil192:
It hides the backgrounds (e.g. ruled paper) of the notes on the home screen

ceskyDJ:
> It hides the backgrounds (e.g. ruled paper) of the notes on the home screen

Ok, I see. I tried only image background before and nothing was changed after changing the option in settings :).

PR #390, opened 2023-02-22
i18n: Add French

Polux49:


adil192:
Thank you!

adil192:
Would you be interested in being mentioned when there's new strings that need to be translated? https://github.com/adil192/saber/issues/19#issuecomment-1306096416

PR #398, opened 2023-02-22
 i18n: Update Brazilian Portuguese

diego-mrqs:
After testing the app with the Brazilian Portuguese translation (v0.9.4), I made these minor improvements.


adil192:
Thanks!

PR #416, opened 2023-02-25
i18n: Update French

Polux49:
I don't understand the @@info instruction on [_missing_translations.json](https://github.com/adil192/saber/blob/main/lib/i18n/_missing_translations.json):
"After editing this file, you can run 'flutter pub run slang apply --locale=<locale>' to quickly apply the newly added translations."
Apply it to... what? 
Instead I have edited the JSON file online, I hope it's OK.

### DIFF
--- a/lib/i18n/_missing_translations.json
+++ b/lib/i18n/_missing_translations.json
@@ -51,10 +51,10 @@
   "fr": {
     "settings": {
       "prefLabels": {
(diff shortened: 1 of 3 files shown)


adil192:
You can ignore that don't worry!
But if you're interested, it applies the missing translations to the actual language file (`lib/i18n/strings_fr.i18n.json`). I typically prefer people don't do this in pull requests since it makes the diff a little harder to read and I run the commands anyway

PR #491, opened 2023-03-20
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Thanks for image! I had no idea about this meaning of English word "staffs" :D. It is completely different in Czech.

ceskyDJ:
Maybe I'm blind but I don't see button for settings as draft. So, don't merge it, please. I have an idea for improving translation of other background patterns.

ceskyDJ:
Ok, so now it is completed.

PR #510, opened 2023-03-27
stroke: function to calculate straightness of a line

adil192:


WasteOfO2:
Not sure if i am getting the explaination right, but perhaps this approach could be problematic with closed figures?

Closed Polygons have same (x,y) coordinates at the start and the end(ideally). This can perhaps lead to problems where a figure is being treated as a point, rather than a shape. I assume things get more complicated as we start adding more sides to the figure. Perhaps symmetrical polygons beyond certain number of sides could be treated more like circles, which would be trivial to detect.

I see that u are looking to add detection for conics. I am not sure how to implement standard equation forms of conics into code, but perhaps u being a math student helps :P

WasteOfO2:
It could be possible to detect closed figures, provided angle between any 2 lines could be found out. This _can_ theoretically be done by trying to detect a bunch right angle triangles between 2 end points, using sine or cosine laws, or simple trignometric ratios to determine inverse of sine and cosine to find out angles.

I.E, the second approach would be to consider various diagonals on the polygon and calculating 'theta'

Screenshot_20230328-194426_Saber

If such angles could be found between all lines, they will universally add up 2π radians or 360° (whatevs the code accepts)

Sine and cosine laws are applicable to _any_ triangle, meaning if the algorithm could detect 2 end points of a line segment having 1 common point, it could join those 2 lines and again solve for 'theta'

Screenshot_20230328-195548_Saber

Length of linesegments can be determined using distance between 2 points formula, let (x1, y1) and (x2, y2) be 2 points, then sqrt[ (x1-x2)^2 + (y1-y2)^2 ] will give the dist between these points

Same principle of angles adding upto 2π radians or 360° can then be used. 

If summation of all such angles 'theta' add upto 2π radians or 360° (or somewhere close), then we say it is a close figure, no straight line detected.

All this logic **only** proceeds if starting and ending points are found be same (or somewhat close). This _should_ cover all edge cases.

Obv, the part determining triangles is the hard part, I have covered the required trignometric and geometric logic here.

I am not sure how feasible it is to implement this, but I am confident this should rememdy the issue or give some idea how to tackle it

WasteOfO2:
However, closed conics like circles and ellipses dont follow all that stuff mentioned above. We need another way to determine those

Here, parametric forms of circle and ellipse are helpful, but an easier approach for circles exists

Screenshot_20230328-202359_Saber

Points A, B and C always satisfy for equation S. But by finding the distance between G (the centre) and any of the points A, B and C, we can find the radius of the circle. By taking multiple such measurements, find average of the radii, we can find if the distances between AG, BG and CG are relatively same and consider it as a circle. Multiple such points along the circumference could be considered to more accurately determine whether or not it is a circle

Or by putting coordinates of A, B and C in x and y in the standard equation for a circle, we can see if the points are concyclic.

I am not a 100% sure for ellipses, will update if i can find a viable solution for them

PR #513, opened 2023-03-28
i18n: Update Turkish

fybx:
Sorry, I couldn't generate `strings.g.dart`, but here are the translated strings 💌


adil192:
Don't worry about generating `strings.g.dart`. The PR is easier to read without it, plus I need to run the same commands either way.

adil192:
Thanks!

PR #551, opened 2023-04-08
i18n: Update Turkish

fybx:
Are we supposed to keep `(OUTDATED)` in the key?


adil192:
> Are we supposed to keep `(OUTDATED)` in the key?

It doesn't matter either way :)

adil192:
Thanks!

Issue #553, opened 2023-04-09
Layering features of the notes (Support of Cornell's Note taking method)


akib-89:
### Motivation

I use Cornell's note taking method for my notes. so when revising the notes some portion of the notes should be covered. If there is a good layering is done I can easily turn of the visibility of a single layer and test form my memory. 

### Solution

There should be some way to divide the note page in 4 sections. (heading, main words, brief description, Summary) this can be done adding another template or giving the user the ability to put lines across the notes. then the brief description sections should be written on a separate layer. whenever the user press a button the layer becomes invisible and with the press of a button the layer becomes visible again. The positioning of the button can be done in the toolbar currently present.

### Saber version

v0.11.2 Google Play Store

### Device

- Device: [Xiaomi pad 5]
- OS: [Android 13]

### Anything else?

There is an app called "touch notes" that implements the Cornell's system pretty well but the features are locked behind a paywall. It would be really great to see  a open source community outperforming the app. The google play link to the app: [Touchnotes](https://play.google.com/store/apps/details?id=com.newskyer.draw&hl=en&gl=US)

PR #554, opened 2023-04-09
patterns: add Cornell notes pattern

adil192:
- Loosely related to #553 

image


adil192:
@akib-89 Please can I have your feedback on a Cornell notes template?
Would you prefer the boxes around each section? And would you prefer each section having a title/subheading?

akib-89:
Yes this looks good to me. But is it possible to provide any process of making this section from the user end not by the template. Template is fine though.

adil192:
> But is it possible to provide any process of making this section from the user end not by the template.

I suppose you'd be able to draw the lines yourself, but why do you ask?

akib-89:
Its completely fine. Thank you.

PR #574, opened 2023-04-10
i18n: Update Czech

ceskyDJ:


ceskyDJ:
@adil192, I'm not sure about Flathub metadata file translation. I found only https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Localization.html, which works completely different.

ceskyDJ:
Btw, we need to translate:

- name
- summary
- description

ceskyDJ:
I found it here, btw: https://docs.flatpak.org/en/latest/conventions.html.

ceskyDJ:
I added translation of desktop entry file, too (see https://www.baeldung.com/linux/desktop-entry-files).

adil192:
> @adil192, I'm not sure about Flathub metadata file translation. I found only https://www.freedesktop.org/software/appstream/docs/sect-Metadata-Localization.html, which works completely different.

That's for separately installable language packs, whereas Saber just includes all the languages.

adil192:
> I found it here, btw: https://docs.flatpak.org/en/latest/conventions.html.

Which part of this do you want me to look at?

adil192:
> I added translation of desktop entry file, too (see https://www.baeldung.com/linux/desktop-entry-files).

Thanks, I'll add this to the README

ceskyDJ:
> > I found it here, btw: https://docs.flatpak.org/en/latest/conventions.html.
> 
> Which part of this do you want me to look at?

If it is just for separated language packs, it doesn't matter.

ceskyDJ:
> Btw, we need to translate:
> 
> - name
> - summary
> - description

@adil192, what about other items of this list? Is it possible to just use the `language` attribute for them, too?

adil192:
> Btw, we need to translate:
> 
>     * name
> 
>     * summary
> 
>     * description

I can't see anything in the AppStream spec about translating the name and summary.
I think (at least for now) we can just translate the description tag and also the `.desktop` Comment as already written in the [instructions](https://github.com/adil192/saber#translating-app-store-descriptions)

adil192:
@ceskyDJ is this ready to merge?

ceskyDJ:
Oh, I'm sorry. I forgot about a little. There is missing Flatpak translation, I think. So, you can copy translation from elsewhere and merge it.

PR #598, opened 2023-04-24
add inlang to make the contribution of translations easier

NilsJacobsen:
Let me know what changes you request for merging this PR.

### Description
This pull request adds the possibility for contributors and translators to manage translations in a UI instead of files with no overhead for the maintainers.

To get a UI for translations contained in this repository, an [inlang.config.js](https://inlang.com/documentation/config) has been created at the root of the repository. Furthermore, the translation files have been cleaned and I added some missing german translations.

### Preview
The changes of this PR and a live instance of the editor can be previewed with the following link https://inlang.com/editor/github.com/NilsJacobsen/saber/. Note: This link should be changed to point to saber instead of NilsJacobsen after this PR has been merged.

### Badge
If needed you can use also a dynamic badge, which is going to show you the current status. You can see a demo badge rendered based on my fork.
[translation badge](https://inlang.com/editor/github.com/NilsJacobsen/saber?ref=badge)

### Limitations
**Certain actions are slow**
Inlang is running entirely on git, giving tremendous CI/CD and contribution power to localization (and potentially other verticals -> [why inlang](https://inlang.com/documentation/why-inlang)). That means that the whole saber repo is cloned into the browser which makes certain actions like the initial load and pushing changes slow. Those limitations will be fixed with future releases and require no input from saber.

In order to make this PR work i moved the english reference language in the crowd-sourced community folder. So to make that change ready for production the pointer to the new location needs to be relocated. 

If you want to use features like machine translate it makes sense to name the different files like this `en.json` instead of `strings_en.i18n.json` but this is no hard requirement. 

pika-1680686287719-1x

Preview the messages on https://inlang.com/editor/github.com/NilsJacobsen/saber.


adil192:
Hi, thanks for your thorough PR :)
Please can you explain why you chose inlang over other similar softwares?

adil192:
>In order to make this PR work i moved the english reference language in the crowd-sourced community folder. So to make that change ready for production the pointer to the new location needs to be relocated.

I think it would be better to use a symlink if possible

adil192:
> If you want to use features like machine translate it makes sense to name the different files like this `en.json` instead of `strings_en.i18n.json` but this is no hard requirement.

The `slang` filename format is `<namespace>_<locale?>.<extension>` so this won't be possible.

> Furthermore, the translation files have been cleaned and I added some missing german translations.

This would break some functionality as you've removed the [lists](https://pub.dev/packages/slang#-lists) feature of slang. We would require some refactoring to change from lists to dicts but I still think lists are better to use in these scenarios.

adil192:
It also doesn't seem to respond to [_missing_translations.json](https://github.com/adil192/saber/blob/main/lib/i18n/_missing_translations.json) which is what we're currently using.
Namely, the missing feature is responding to the `OUTDATED` flag (and also lists as above).

adil192:
Following this, I don't think inlang would be a good fit for Saber. I'd want something more strongly integrated with [`slang`](https://pub.dev/packages/slang) and its features

NilsJacobsen:
HI @adil192, thanks for your quick rely.

> Hi, thanks for your thorough PR :) Please can you explain why you chose inlang over other similar softwares?

The key thought is to open up the translations to any person, even tech, non-tech or beginners. To make it as easy as possible to contribute translations, BUT and that is important, we want to make it as easy as possible for projects to use inlang without changes in the project. Basically the config is the only thing that you need to add to use the editor, and also the rest of the infrastructure, like vscode extension, sdk ... We do that by using read and right functions that are imported from plugins. Inlang has the following plugins right mow: https://github.com/inlang/ecosystem

> It also doesn't seem to respond to [_missing_translations.json](https://github.com/adil192/saber/blob/main/lib/i18n/_missing_translations.json) which is what we're currently using. Namely, the missing feature is responding to the `OUTDATED` flag (and also lists as above).

Right, inlang does not support that. But it supports linting, so you could easily access lint reports and get a list of missing translations for example.

I see that you want to stick to slang. And even though we are currently working on our own SDKs, that is absolutely not a bummer. If you want to stick to the file naming we can help you be adjusting the json plugin for your needs. If that's what is holding u back I could support with that?

Issue #628, opened 2023-05-05
Some (rasterised) PDFs don't invert properly


adil192:
### Describe the bug

Some PDFs invert to a completely black rectangle. Other PDFs invert weirdly -- showing outlines of letters and not their fill.

### To reproduce

1. Import a PDF with text in it
2. Turn on dark mode and inversion
3. See result like screenshots 

### Expected behavior

Invert normally

### Saber version

v0.12.0 (12010)

### Device

- Device: Google Pixel 6 Pro
- OS: Android 13

### Anything else?

<img src="https://user-images.githubusercontent.com/21128619/236485795-d2b7d460-9a42-497b-a7bc-2ad24eea7db7.png" width=300><img src="https://user-images.githubusercontent.com/21128619/236485612-3f94d487-395c-41ca-aaac-6d7452030219.png" width=300> <img src="https://user-images.githubusercontent.com/21128619/236486695-315ca784-d47e-422a-be30-cc7948910e51.png" width=300>

- May be solved by #627 
- #432 might cover up this issue but it will remain for PDFs imported before #432 is completed

adil192:
Both of these problems seem to be fixed in v0.12.9, maybe because of upstream flutter fixes or because of https://github.com/adil192/saber/commit/8fe64c24b1a32fa65090f32c92df0fa0fa8d45d5.

adil192:
Nope they're only fixed on my Galaxy Tab S7 (OneUI 5.1/Android 13) but not my Pixel 6 (Android 13)

adil192:
Testing image:

| Before | After |
| --- | --- |
| ![Before: Only the outlines of shapes are shown](https://github.com/saber-notes/saber/assets/21128619/c9c8ed49-0e88-40fc-bf8d-b3191682cc38) | ![After: The outlines and fill of the shapes are correctly shown](https://github.com/saber-notes/saber/assets/21128619/f338a6f0-5a09-4576-a705-406f46459d70) |

PR #632, opened 2023-05-08
invert: use highest filter quality

adil192:


adil192:
- This was intended to fix #628 but didn't make any difference

Issue #568, opened 2023-04-10
Improve Saber description in app stores


fischerjosue:
### Motivation

I think Saber app description in app stores is too short and needs so much more information about what the app can do.

### Solution

Improve the Saber description to include everything the app can do so far.

This may be an example of the app description:

Saber is an innovative handwritten note-taking app that combines the convenience of handwriting with the versatility of digital technology. With Saber, users can handwrite their notes on the screen of their device, whether on a tablet, smartphone or computer, and even draw using a digital pen or their finger, offering a smooth writing experience. And natural, similar to writing on paper.

Saber allows users to put their ideas down, draw diagrams or schematics, make sketches, or simply take notes in their own handwriting. The app offers a wide range of pen options, including different colors, thicknesses, and styles, as well as highlighters, allowing users to personalize their notes and make them visually appealing.

Another advantage of Saber is its cross-platform synchronization in real time. Users can access their notes from multiple devices and platforms, allowing them to work on their notes on their tablet during a meeting and then access them on their smartphone on the way home. This provides great flexibility and portability in the way users interact with their notes.

Capture your ideas with the freedom to write by hand anytime, anywhere with Saber: The best way to bring your creativity to the digital world!

### Saber version

v0.11.3

fybx:
+++ this needs to be done. I also think Saber is ready to publish as a complete app that will improve with every update.

licaon-kter:
Add more here: https://github.com/adil192/saber/tree/main/metadata ?

fybx:
> Add more here: https://github.com/adil192/saber/tree/main/metadata ?

Done 👍🏻

ceskyDJ:
It looks good but you forgot about some features:
- whiteboard
- special backgrounds like stuff, Cornell, grid, etc.
- images and text combination with handwriting
- translations to so many languages by native speakers
- so many configuration options

I think there should be a note about NextCloud support, too.

iamjosuefischer:
Saber is an innovative cross-platform handwritten note-taking app, available for Android, iOS, Linux, Windows, and macOS. That combines the convenience of handwriting with the versatility of digital technology. With Saber, users can handwrite their notes on the screen of their device, whether on a tablet, smartphone, or computer and even draw using a digital pen or their finger, offering a smooth writing experience. And natural, similar to writing on paper. Saber is available in so many languages, translated by native speakers.

[...]

PR #694, opened 2023-06-09

jhihyulin:


jhihyulin:
In the meantime I noticed #568 issue, is there a plan to implement it.

ceskyDJ:
> In the meantime I noticed #568 issue, is there a plan to implement it.

It doesn't work properly in Flathub (see https://flathub.org/apps/com.adilhanney.saber, all versions are displayed every time). But I think at least somewhere (.desktop file) it should work well. So, thanks for translating!

Issue #678, opened 2023-05-31
Add a laser pointer for presentations while using the screen recorder


m0byn:
### Motivation

Additionally to a highlighter and pens a laser pointer just showing a colored dot on the screen would be of much use while using the screen recorder and presenting online.

### Solution

As far as I know GoodNotes has a functionality where the pointer is shown on the screen, and if you write anything, it vanishes after some seconds when lifting the pen. For illustration see the following [video](https://youtu.be/c8pWXZp0FC8) at 1:30. This illustrates exactly what I am looking for!

### Saber version

v0.12.9

### Device

- Device: Samsung Galaxy Tab S6
- OS: LineageOS 20 (Android 13)

m0byn:
I do have a recommendation for the current laser pointer tool. 
As it appears to me, the lines written by the laser pointer start vanishing x seconds after drawing them. The problem arises with the fact that every touch of the display erases x seconds after drawing, and not the full line deletes after x seconds of lifting the pen. So, if I keep the pen on the screen to display the written lines longer, then the last bit of line takes forever to vanish. See the following video for illustration 

https://github.com/adil192/saber/assets/33026003/dcab19ad-6b59-4a84-b499-93e610ed80ea

Solution: let the complete line vanish x seconds after lifting the pen. 

Let me know if I shall open a new issue instead of commenting here!

PR #716, opened 2023-06-16
i18n: Update Czech

ceskyDJ:
Oh, you got me, @adil192! What do you mean by "layout size"? I think I couldn't translate it as "size" to Czech, because it won't be understandable. So, I translate it as "User interface layout (type)": "Auto/Phone/Tablet". Are you ok with this change?

And the laser pointer... Is it real pointer? Why is it in the pens menu? I mean something for pointing (without drawing) around my note, when I streaming/showing the note to some people.

If you are ok with my little edit of the layout size and the laser pointer is really a pointer accepted by by definition, you can merge this pull request. For future translations, could you add screenshot or more detailed description, please? It is sometimes very hard to understand what you mean, when I didn't see it and I haven't got any deeper explanation.

### DIFF
--- a/lib/i18n/_missing_translations.json
+++ b/lib/i18n/_missing_translations.json
@@ -65,17 +65,17 @@
   "cs": {
     "settings": {
       "prefLabels": {
-        "layoutSize": "Layout size"
+        "layoutSize": "Rozložení uživatelského rozhraní"
       },
       "layoutSizes": {
-        "auto": "Auto",
-        "phone": "Phone",
+        "auto": "Automatické",
+        "phone": "Telefon",
         "tablet": "Tablet"
       }
     },
     "editor": {
       "pens": {
-        "laserPointer": "Laser pointer"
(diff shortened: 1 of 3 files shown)


adil192:
> So, I translate it as "User interface layout (type)": "Auto/Phone/Tablet". Are you ok with this change?

That's perfect, I'll probably change the English translation to "Layout type" too.

> And the laser pointer... Is it real pointer? Why is it in the pens menu? I mean something for pointing (without drawing) around my note, when I streaming/showing the note to some people.

The laser pointer draws lines that fade out after 1s, just like the video you linked in #678. It's in the pens menu because internally it's implemented in the same way as a pen.

> For future translations, could you add screenshot or more detailed description, please? It is sometimes very hard to understand what you mean, when I didn't see it and I haven't got any deeper explanation.

Yes I'll do that sorry

ceskyDJ:
Ok, thanks! So, you're free to merge it.

PR #726, opened 2023-06-18
snap: explicitly set version

soumyaDghosh:
In my repo, the version was adopted from the appstream metadata file which is also used by the flatpak. But, after shifting to this repo, I noticed that for the last two versions, it's still showing the version 0.12.10. So, I am changing this manually. The github workflow will also change according to this, which I will also do.


adil192:
Thanks

PR #731, opened 2023-06-19
i18n: Update Czech

ceskyDJ:
I'm glad to see Saber finally has folders! Nice work, @adil192! Thank you very much!


adil192:
> I'm glad to see Saber finally has folders! Nice work, @adil192! Thank you very much!

Not just yet but we're getting there!

adil192:
Thanks!

PR #732, opened 2023-06-19
i18n: Update Russian

ShevT:


adil192:
Please avoid editing `strings_ru.i18n.json` unless you need to edit existing translations

adil192:
Thanks!

ShevT:
> Please avoid editing `strings_ru.i18n.json` unless you need to edit existing translations

Those, is it enough to edit the corresponding section of the _missing_translations.json file?

adil192:
Yep!

Issue #733, opened 2023-06-19
Snap versioning not working


soumyaDghosh:
### Describe the bug

So, as said before, after the migration, the versioning of the snap somehow got messed, and it's still showing the version as `0.12.10`. 

### To reproduce

try by installing the snap from any channel

### Expected behavior

It should have shown the proper version

### Saber version

0.12.10 Snap

### Device

- Device: Desktop Linux
- OS: Any

### Anything else?

For now, Users please look over it, the updates are given properly, which you can find within the app itself. We trying to fix it asap.

Keep any eye to this:
https://forum.snapcraft.io/t/versioning-not-working-properly-with-snap-saber/35677

soumyaDghosh:
Wait, let me have some local tests.

adil192:
Please don't pin issues. You only have collaborator permissions to link the Snap to this repo

soumyaDghosh:
Sorry about that.

PR #734, opened 2023-06-19
snap: fix versioning

soumyaDghosh:
This is the temporary fix for #733
I will still look into that.

### DIFF
--- a/flatpak/com.adilhanney.saber.metainfo.xml
+++ b/flatpak/com.adilhanney.saber.metainfo.xml
@@ -66,7 +66,8 @@
         </screenshot>
     </screenshots>
 
-        <release version="0.13.1" type="development" date="2023-06-18">
+    <releases>
+         <release version="0.13.1" type="development" date="2023-06-18">
             <description>
                 <ul>
                     <li>Checks that the whiteboard is saved before you navigate away</li>
@@ -87,7 +88,6 @@
                 </ul>
             </description>
         </release>
-    <releases>
         <release version="0.12.10" type="development" date="2023-06-06">
             <description>
                 <ul>
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,12 +1,11 @@
 name: saber # you probably want to 'snapcraft register <name>'
 base: core22 # the base snap is the execution environment for this snap
 adopt-info: saber
-version: 0.13.1
 grade: stable # must be 'stable' to release into candidate/stable channels
 confinement: strict # use 'strict' once you have the right plugs and slots
 license: GPL-3.0
 compression: lzo
-icon: assets/icon/icon.png
+icon: snap/assets/icon/icon.png
 donation: https://github.com/sponsors/adil192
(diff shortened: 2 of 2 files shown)


snap/snapcraft.yaml
```diff
@@ -1,12 +1,11 @@
 name: saber # you probably want to 'snapcraft register <name>'
 base: core22 # the base snap is the execution environment for this snap
 adopt-info: saber
-version: 0.13.1
```
adil192, line 4: We should probably leave this in (but with single quotes) anyway

snap/snapcraft.yaml
```diff
@@ -1,12 +1,11 @@
...
 grade: stable # must be 'stable' to release into candidate/stable channels
 confinement: strict # use 'strict' once you have the right plugs and slots
 license: GPL-3.0
 compression: lzo
-icon: assets/icon/icon.png
+icon: snap/assets/icon/icon.png
```
adil192, line 8: Is this the right path?

snap/snapcraft.yaml
```diff
@@ -36,7 +35,7 @@ parts:
     after: [rustup]
     source: https://github.com/adil192/saber.git
     source-type: git
-    source-tag: v$SNAPCRAFT_PROJECT_VERSION
+    source-tag: v0.13.1
```
adil192, line 38: And leave this as it was

adil192 (changes requested):

snap/snapcraft.yaml
```diff
@@ -55,6 +54,9 @@ parts:
       - CARGO_HOME: $CRAFT_STAGE/usr/share/rust
       - CARGO_BUILD_JOBS: $CRAFT_PARALLEL_BUILD_COUNT
     override-build: |
+      cd $CRAFT_PART_SRC
+      craftctl set version=$(git describe --tags --abbrev=0 | cut -c 2-)
```
adil192, line 58: Can we use `$SNAPCRAFT_PROJECT_VERSION` here instead of git?

flatpak/com.adilhanney.saber.metainfo.xml
```diff
@@ -66,7 +66,8 @@
         </screenshot>
     </screenshots>
 
-        <release version="0.13.1" type="development" date="2023-06-18">
+    <releases>
+         <release version="0.13.1" type="development" date="2023-06-18">
```
adil192, line 70: Thanks for spotting this! Just to keep the tree clean, please can you revert this commit and I'll re-commit it to the main branch.

adil192: Update: recommited in https://github.com/adil192/saber/commit/916db979c5b05221bdcccb340699137af91ed6ac

soumyaDghosh: Sure, this commit is most probably of no use then. Deleting this PR.

adil192: Why did you delete the whole PR?

PR #735, opened 2023-06-19
i18n: Update German

theIuser:


adil192:
Thanks!
Would you be interested in being mentioned when there's new strings that need to be translated? https://github.com/adil192/saber/issues/19#issuecomment-1306096416

PR #737, opened 2023-06-20
snap: revert back to `source-tag`

soumyaDghosh:
#733


soumyaDghosh:
`libsecret-1-dev` is in the gnome-sdk, so, no need to keep it.

adil192:
> Can you push a hotfix release?

Is there no way to trigger another build from the snap website? I'm not ready to release just yet

soumyaDghosh:
Then we'll have to use the latest commit from the latest tag. Or, I can do something else. Creating a branch, until the last release commit, applying the fix to it, as it's in the later commits(which is also the reason that until a new release, it'll not found the fixed metadata file). And build the snap from there.

adil192:
Yeah sure the branch method seems good

PR #751, opened 2023-06-26
i18n: Update Turkish

fybx:
I have updated several missing keys but left out the ones marked as outdated. What should I do about them? Can I translate, then delete "(OUTDATED)", or should I follow another procedure?

### DIFF
--- a/lib/i18n/_missing_translations.json
+++ b/lib/i18n/_missing_translations.json
@@ -414,86 +414,34 @@
   "ru": {},
   "tr": {
     "home": {
-      "newFolder": {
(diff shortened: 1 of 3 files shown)


adil192:
> Can I translate, then delete "(OUTDATED)", or should I follow another procedure?

I don't think it makes any difference if you delete the `(OUTDATED)` annotation or not. Please translate the strings as normal and you can leave the annotation there

adil192:
I'm going to merge this for now to avoid merge complications, and please can you pick up anything missing in another PR

PR #755, opened 2023-06-28
i18n: Update Czech

ceskyDJ:
@adil192, there is one thing I noticed. You named a field as "Note name", but I think better is "New note name" (so it's more clear). I've already added the "new" to the Czech translation.


adil192:
I think I prefer just "Note name"

adil192:
Thanks!

PR #779, opened 2023-07-06
i18n: Update Czech

ceskyDJ:


lib/i18n/_missing_translations.json
```diff
@@ -29,13 +29,13 @@
     "settings": {
       "prefCategories": {
         "editor(OUTDATED)": "Editor",
```
adil192, line 31: Just to double check, is this meant to be "Editor"?

ceskyDJ: Oh yes, we use the word "Editor" in some contexts just like English does. It's just one of the loanwords we got from English. Machine translation unfortunately returned completely bad string in this case.

ceskyDJ: "Čeština" is a Czech name of the official Czech language. So, it's completely out.

ceskyDJ: It's a little funny, we translate "edit" to "změnit", "editing" to "mění se" and "editor" (e.g. "text editor") to "editor" ("textový editor" in a previous example). It's from old times (for us young people), because "editor" was used for people who edit content of literature (but maybe it started with the newspaper).

PR #783, opened 2023-07-07
feat: add s-pen button erasing

ZebraVogel94349:
This adds the feature to switch to the eraser while the button on a stylus (tested with the Samsung S-Pen) is pressed and to switch back to the previous tool when the button is released, which is similar to the behavior in Samsung Notes.

### DIFF
--- a/lib/components/canvas/canvas_gesture_detector.dart
+++ b/lib/components/canvas/canvas_gesture_detector.dart
@@ -23,6 +23,7 @@ class CanvasGestureDetector extends StatefulWidget {
     required this.onDrawUpdate,
     required this.onDrawEnd,
     required this.onPressureChanged,
+    required this.onStylusButtonChanged,
 
     required this.undo,
     required this.redo,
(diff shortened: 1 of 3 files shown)


adil192:
@ZebraVogel94349 Thanks for this! Would you be able to test the new changes (mostly the part where the user is already drawing a stroke and presses the eraser button)?

ZebraVogel94349:
@adil192 I tested all new changes with my Tab S7 FE and the S-Pen and it's working fine. If a the user is already drawing a stroke and presses the eraser button, the currently drawn stroke gets removed and the eraser is selected. The only problem is, that the button press gets only detected, when the pen is already touching the screen, but not when hovering. I tried using the onPointerHover event for that, which worked for switching the tools, but you can't draw anything when pressing the button and then touching the screen, due to [this issue in flutter](https://github.com/flutter/flutter/issues/42846). I don't know if it only affects Samsung devices or also any other device with a stylus.

adil192:
I'll merge this in now. Hopefully the flutter issue is fixed soonish!
Thanks again :)

TomBPotochek:
I think this change made it so it is now impossible to manually erase with the eraser. When I select the eraser and try to use it, it immediately switches back to the pen tool. Happens with both s pen and finger drawing.

adil192:
> I think this change made it so it is now impossible to manually erase with the eraser. When I select the eraser and try to use it, it immediately switches back to the pen tool. Happens with both s pen and finger drawing.

This should be fixed in #785

PR #786, opened 2023-07-09
i18n: Add Japanese

nagiyu:
Created Japanese translation of lib/i18n/strings.i18n.json.

All parts that can be translated in "strings.i18n.json" have been translated.


adil192:

nagiyu:
Thank you for accepting my pull request. Yes, please.

PR #788, opened 2023-07-09
build the snap directly from main branch for this update

soumyaDghosh:
This is to fix the metadata, as it makes the versioning not working


adil192:
I'm skipping this update for flatpak since the changes are just for Android. Is it okay if we skip this update for snap too?

soumyaDghosh:
Sure, as you wish, the update is done, it's in the edge channel, but the version is '0.14.4'. But, I don't see any problem here TBH

PR #792, opened 2023-07-10
i18n: Update Czech

ceskyDJ:


ceskyDJ:
It's a good move to use YAML for translation files. It's more human friendly, so maybe some non-developer translator could translate Saber, too. I like TOML a little more (it's got a flat design, so there aren't so many indents) but YAML is good, too, for these thing. So, good choice!

Issue #607, opened 2023-04-28
Eraser: switching back to the pen after using


Iey4iej3:
### Motivation

The erasers are usually used for a very short period of time. It would be better if there is a mode of eraser so that it is switched back to the previous state (pen or mark pen, say) after using it.

### Solution

For example, it switches back (to the pen/mark pen, etc) when the stylus leaves the screen.

### Saber version

v0.11.3 F-Droid (11030)

### Device

- Device: Samsung Tab S6 Lite
- OS: Lineage OS 19 (Android 12)
- Browser (if applicable): [e.g. Firefox]

PR #807, opened 2023-07-17
feat: switch back to pen after erasing

ZebraVogel94349:
This fixes [#607](https://github.com/adil192/saber/issues/607) by adding a setting to switch the tool back to the pen after erasing.


adil192 (approved):

adil192:
Thank you!

PR #817, opened 2023-07-19
i18n: Update Czech

ceskyDJ:


ceskyDJ:
@adil192, is it really necessary to run Codecov on pull requests, witch only updates community translations? I think it would be better to ignore translations in case of Codecov reports (It's surely configurable in Github Actions, but I didn't check the way you run that).

adil192:
> @adil192, is it really necessary to run Codecov on pull requests, witch only updates community translations? I think it would be better to ignore translations in case of Codecov reports (It's surely configurable in Github Actions, but I didn't check the way you run that).

There's no way (that I know) to distinguish between a PR that changes coverage or not, since even translation PRs sometimes require changes in other files

ceskyDJ:
Hmm, it should be ignored. interesting.
image

adil192:
That's just for what's counted towards the coverage percent, not if codecov comments on a PR

ceskyDJ:
Ok, I see.

PR #822, opened 2023-07-19
workflows: Don't run tests for pull requests with i18n updates

ceskyDJ:
When the pull request changes only files from lib/i18n, tests won't be run.

### DIFF
--- a/.github/workflows/tests_main
+++ b/.github/workflows/tests_main
@@ -1,0 +1,52 @@
+name: Run main tests
+
+on:
+  # Enable manual run
+  workflow_dispatch:
+  pull_request:
+  push:
+
+env:
+  project-id: com.adilhanney.saber
+
+jobs:
+  run-tests:
+    name: Run Flutter tests
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+
+      - name: Remove unneeded files
+        run: |
+          rm -rf assets_raw
+          rm -rf submodules
+
+      - name: Install Flutter dependencies
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


ceskyDJ:
Source: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-excluding-paths.

ceskyDJ:
So, maybe there is a way. I use something like that in one of my repositories and it work well for me.

adil192:
Thanks :)

adil192 (changes requested):
Wait is there a way to only turn off the codecov part of the workflow? I use some tests to avoid some common mistakes in translations like spelling Saber as "Sabre"

ceskyDJ:
At least you can divide it into more workflows. There are conditions (ifs), too, but I don't know if it supports this. I'll take a look.

ceskyDJ:
Oh, it isn't supported, as I thought. So, I'd choose one of these ways:
- create two workflows in this case – one for code, one for translations,
- or extract the test you need to run always to its own workflow and use my original example.

ceskyDJ:
As you have it prepared for a manual run, too, I'd prefer the first way. I'll commit an example of my thought.

ceskyDJ:
So, what about something like that? Another option is divide just tests for spelling, etc. to own workflow. You'll need to edit conditions for pull requests (checks required to pass) in Github repository settings.

adil192:
The "Upload to Codecov" requires `flutter test --coverage` to be run first to obtain the coverage information, so this wouldn't work as it is. Creating a separate workflow would work, but this seems like too much work just to hide a comment on PRs so I'm going to close this

PR #824, opened 2023-07-20
i18n: Update Simplified Chinese

Jesse205:
update simple chinese


adil192:

Jesse205:
Yes, I would be happy to help translate the software into Simple Chinese.

PR #844, opened 2023-07-24
Update _missing_translations.yaml

Jesse205:
update simple chinese


ceskyDJ:

adil192:
Thanks!

PR #845, opened 2023-07-24
i18n: Update Russian

ShevT:


ceskyDJ:

ShevT:
> don't forget to translate app store description (and optionally README, too)! 

How to do it?

ceskyDJ:
It's described in README, see https://github.com/adil192/saber#translating-app-store-descriptions. Is it enough or do you want some more information? In case you need some, let me know. Alternatively see #846 or #843.

ceskyDJ:
@adil192, could you add Flathub description? We could try more languages than just English and Czech, after the content of the Flathub website will be updated.

PR #846, opened 2023-07-24
i18n: Update Czech

ceskyDJ:


ceskyDJ:
So, here a Czech translation of your new app description. It's translated a little more "freely", so it could sound better (more native for Czech people). I've translated README, too. There are still some things (like images) in English, but the text should be fully translated.

ceskyDJ:
I don't know if we want to fully translate README files (some sections like "Translating" maybe doesn't make sense), so you could do anything you want with it :D.

ceskyDJ:
I'm not sure about that :D. For now, I've translated it completely, but I think some parts isn't necessary. Maybe let's open a discussion for translators and read their opinions on translating README. What do you think about it?

adil192:
We could probably leave out the `Translating` and `Development Notes` sections to avoid it being too technical for our translators. 
I'm not too sure about opening a discussion for translators since I haven't gotten their consent for that.

ceskyDJ:
Ok. I found some issues in Czech README, so I'll update it with the fix.

Issue #196, opened 2023-01-01
Open .sbn files


FBarrca:
**Is your feature request related to a problem? Please describe.**
Cant open .sbn fles
**Describe the solution you'd like**
Click on a .sbn file in the file explorer to open the file.

**Device (please complete the following information):**

 - OS: Android 13 and Windows 11
 - Version: [e.g. v0.5.2.0]

**Additional context**
On Windows, clicking a file opens the app, but it doesnt open the note.
On Android, the app doesnt appear as an option to open the file with. 
I've been looking a bit into the Android side and this might be the solution to enable open with.
https://pub.dev/packages/receive_sharing_intent

Hope it helps!

adil192:
I've been looking for a flutter package to do that, thanks!
I'll probably go with a similar package https://pub.dev/packages/flutter_sharing_intent as it has better cross-platform support.

Linking discussion https://github.com/adil192/saber/discussions/163

FranciscoBarCa:
A 4 day old package. How lucky! 
Although I think it is misslabeled in the site, that package only supports IOS atm.

adil192:
> Although I think it is misslabeled in the site, that package only supports IOS atm.

How do you know?

adil192:
> Although I think it is misslabeled in the site, that package only supports IOS atm.

Looks like it supports Android and iOS but nothing else yet: https://github.com/bhagat-techind/flutter_sharing_intent/issues/1#issuecomment-1361227715

adil192:
[flutter_sharing_intent](https://pub.dev/packages/flutter_sharing_intent) doesn't support opening json files (yet) so I'm switching to [receive_sharing_intent](https://pub.dev/packages/receive_sharing_intent). I'll need to change this in the future as [receive_sharing_intent](https://pub.dev/packages/receive_sharing_intent) is unmaintained and only supports Android and iOS.

adil192:
This likely only works for Android so I'll leave this issue open. We're waiting on [flutter_sharing_intent](https://pub.dev/packages/flutter_sharing_intent) to be further developed

adil192:
> @adil192 Hi, I am using AOSP , its file manager does not seems to open in saber. I tried share option Saber is not appearing in list. Could you elaborate How to import file in saber? 

Saber should come up in the share menu. Which Android version and Saber version are you on?

Luciogi:
@adil192 I am using Android 13 and Saber 0.5.3

adil192:
[...]

PR #863, opened 2023-08-04
feat: importing .sbn and .sbn2 files

ZebraVogel94349:
This replaces the "+" button in the home menu by a button which opens a little menu where you can choose whether to create a new note or to import a note. When choosing the import option, the file picker opens and you can select a .sbn or .sbn2 file which will be imported. This addition makes it possible to open note files on Android, which was not easily possible before as mentioned in [#196](https://github.com/adil192/saber/issues/196).


lib/pages/home/browse.dart
```diff
@@ -168,18 +169,59 @@ class _BrowsePageState extends State<BrowsePage> {
           ],
         ),
       ),
-      floatingActionButton: FloatingActionButton(
-        shape: cupertino ? const CircleBorder() : null,
-        onPressed: () async {
-          final newFilePath = await FileManager.newFilePath('${path ?? ''}/');
-          if (!mounted) return;
-          context.push(RoutePaths.editFilePath(newFilePath));
+      floatingActionButton: SpeedDial(
```
adil192, line 172: Please can you make this a separate widget so the code isn't duplicated in browse.dart and recent_notes.dart?

adil192 (changes requested):

adil192 (approved):

adil192:
Thank you for your contributions!
Would you be willing to submit another PR which allows importing PDFs with the same "Import note" button?

PR #869, opened 2023-08-07
feat: import pdf from home menu

ZebraVogel94349:
This makes it possible to import PDF files when selecting the "import note" option in the home menu and then selecting a PDF file in the file picker. This should also <s>fix #[771](https://github.com/adil192/saber/issues/771)</s> by setting the "withData" parameter in the file picker to false.

### DIFF
--- a/lib/components/home/new_note_button.dart
+++ b/lib/components/home/new_note_button.dart
@@ -2,6 +2,7 @@ import 'package:file_picker/file_picker.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_speed_dial/flutter_speed_dial.dart';
 import 'package:go_router/go_router.dart';
+import 'package:printing/printing.dart';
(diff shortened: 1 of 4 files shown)


adil192:
I don't think this would solve https://github.com/adil192/saber/issues/771 since the full content of the PDF is still loaded into memory

ZebraVogel94349:
> I don't think this would solve https://github.com/adil192/saber/issues/771 since the full content of the PDF is still loaded into memory.
When I tried importing large .sbn2 files with the parameter set to true, the app always crashed on my tablet. With the parameter set to false, it didn't crash anymore. According to https://github.com/miguelpruivo/flutter_file_picker/issues/392 the files are not loaded into memory directly after picking them, but only cached, if withData is set to false. The content is still loaded into memory, but I think, the problem is not a lack of memory to store the content of a file but the FilePicker Library trying to allocate the memory to quickly/without checking if there is enough memory available.

PR #871, opened 2023-08-09
feat: save points as BsonBinary

ZebraVogel94349:
This changes the way points are stored in BSON files. Previously, they were stored using the double format in BSON (8 Bytes per value, so 24 for each point (16, if the pressure is not stored)). When converting them to a Float32List and storing that into a BsonBinary value, this gets reduced to 4 Bytes per value, so 12 for each point (8 without pressure) and also now the names of these values 'x', 'y' and 'p' are not stored anymore, which should reduce the file size further. In total, this reduces the file size of a standard handwritten note by about 50%. Old notes can still be opened normally.

### DIFF
--- a/lib/components/canvas/_stroke.dart
+++ b/lib/components/canvas/_stroke.dart
@@ -71,7 +71,7 @@ class Stroke {
     required this.penType,
   });
 
-  Stroke.fromJson(Map<String, dynamic> json) :
+  Stroke.fromJson(Map<String, dynamic> json, int fileVersion) :
         _isComplete = json['f'],
         pageIndex = json['i'] ?? 0,
         penType = json['ty'] ?? (Pen).toString() {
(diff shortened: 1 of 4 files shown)


ceskyDJ:
It looks like a good work!

Btw, float is just a half precision data type, so isn't that a problem?

ZebraVogel94349:
> Btw, float is just a half precision data type, so isn't that a problem?

The precision of a Float32 is about 8 significant digits, which is more than enough and shouldn't result in any noticable loss of quality. On bigger pages with a lot higher maximum coordinates it could maybe cause issues, but I don't think other page sizes are planned at the moment and the pages would have to be huge for that problem to occur.

ceskyDJ:
Ok, that's nice. Thanks for providing more details!

adil192:
Thanks!

Issue #861, opened 2023-07-30
Bug when creating a note and a folder with the same name


Parameter15:
### Describe the bug

When creating a folder and a note with the same name, the browser page shows two folders with the same name and doesn't show the note.

### To reproduce

1. Create a note called "test"
2. Create a folder also called "test"
3. Go to the browser page
4. You will see two folders called "test", and the note is not shown

### Expected behavior

See both the note and the folder in the browser page

### Saber version

v014.11 F-Droid

### Device

- OS: LineageOS

PR #873, opened 2023-08-14
fix: allow files and folders with the same name

ZebraVogel94349:
This fixes [#861 ](https://github.com/adil192/saber/issues/861) by checking if an entry for a directory that has the same name as a note already exists before adding such an entry, so that an entry for the note is also created.


adil192:
Thanks!

PR #875, opened 2023-08-15
Fix linux-arm64 build

### LINKED ISSUE #874: Build fails on linux/arm64
### Describe the bug

PDFIUM hardcodes the target architecture 

### To reproduce

Run `flutter build linux` on an aarch64 linux system

### Expected behavior

It successfully compiles

### Saber version

main branch

### Device

- Device: Macbook Pro 14"
- OS: Linux fedora 6.4.4-402.asahi.fc38.aarch64

### Anything else?

[dart_pdf developer said](https://github.com/DavBfr/dart_pdf/issues/1448) to put:
`set(PDFIUM_ARCH "arm64" CACHE STRING "Architecture of pdfium used" FORCE)`
in `linux/CmakeList.txt`
It does indeed wok, and now saber compiles correctly and runs fine on arm64 - although it does also segfault when I try to load a pdf.

We should probably detect the architecture automatically in the build script to avoid this problem in the first place.

nik012003:
Fixes #874 by setting PDFIUM_ARCH to the correct architecture.
I could only test it on arm64, so somebody please test it on linux-x64 too.

### DIFF
--- a/linux/CMakeLists.txt
+++ b/linux/CMakeLists.txt
@@ -34,6 +34,12 @@ if(NOT CMAKE_BUILD_TYPE AND NOT CMAKE_CONFIGURATION_TYPES)
     "Debug" "Profile" "Release")
 endif()
 
+# Compilation setting for dart_pdf. 
+# 
+# This is required to download and link against the correct build of pdfium
+string(REPLACE "linux-" "" TARGET_ARCH ${FLUTTER_TARGET_PLATFORM})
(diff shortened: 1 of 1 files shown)


nik012003:
Update: I'm getting this fixed in `dart_pdf` itself, so there will probably be no need to merge this

nik012003:
Fix merged into `dart_pdf`

PR #881, opened 2023-08-21
fix: save exported files as .sbn2 instead of .sbn

ZebraVogel94349:
When exporting a note by using the "export as: SBN" button in the export menu, it was stored as a .sbn file instead of .sbn2, which made it impossible to reopen it without renaming. This changes the file extension to .sbn2.


adil192:
Thanks!

PR #882, opened 2023-08-21
feat: store images as BsonBinary

ZebraVogel94349:
This changes the way images are stored in .sbn2 files. Previously they were stored as an base64 encoded string. The BsonBinary format is about 25% more space efficient and should also be slighty faster than encoding the images in base64.

### DIFF
--- a/lib/data/editor/editor_core_info.dart
+++ b/lib/data/editor/editor_core_info.dart
@@ -19,7 +19,7 @@ class EditorCoreInfo {
 
   /// The version of the file format.
   /// Increment this if earlier versions of the app can't satisfiably read the file.
-  static const int sbnVersion = 13;
+  static const int sbnVersion = 14;
   bool readOnly = false;
   bool readOnlyBecauseOfVersion = false;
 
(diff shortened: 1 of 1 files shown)


lib/data/editor/editor_core_info.dart
```diff
@@ -380,7 +384,7 @@ class EditorCoreInfo {
       'c': initialPageIndex,
     };
 
-    json['a'] = assets.map((Uint8List asset) => base64Encode(asset)).toList();
+    json['a'] = assets.map((Uint8List asset) => BsonBinary.from(asset.buffer.asUint8List())).toList();
```
adil192, line 387: Is there any reason you went with `asset.buffer.asUint8List()` instead of just `asset`?

ZebraVogel94349:
Sorry for the chaotic pull request, i had some trouble with git.

lib/data/editor/editor_core_info.dart
ZebraVogel94349: No. I only did it like that because it worked like that for saving the strokes. The conversion was redundant.

adil192 (approved):
Looks good, have you tested it?

ZebraVogel94349:
> Looks good, have you tested it?

yes, I tested it and everything works as before.

adil192:
Thanks

PR #885, opened 2023-08-24
fix: corrupted notes with colors

### LINKED ISSUE #884: Notes with color corrupting when opening in 0.15.1
### Describe the bug

Colored strokes don't work in Version 0.15.1. All notes containing colored strokes opened in that Version will get corrupted. I think the reason for that is, that the new BSON file format uses int64 instead of int, which is not compatible with the way colors are stored. `color = Color(json['c'] ?? defaultColor.value);` in stroke_properties.dart raises an exception because of that.

### To reproduce

1. Create a note using colors
2. Close and reopen the note
3. It is corrupted

### Expected behavior

Not corrupting the note

### Saber version

v1.15.1 PlayStore

### Device

- Device: Samsung Galaxy Tab S7 FE
- OS: Android 13

### Anything else?

_No response_

ZebraVogel94349:
This should fix #884 by saving colors as BsonInt instead of int (which gets converted to Int64 for some reason). Notes no longer get corrupted and old notes can still be opened normally.

### DIFF
--- a/lib/data/tools/stroke_properties.dart
+++ b/lib/data/tools/stroke_properties.dart
@@ -1,3 +1,4 @@
+import 'package:bson/bson.dart';
 import 'package:flutter/material.dart';
 import 'package:saber/data/tools/highlighter.dart';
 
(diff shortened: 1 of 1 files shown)


ZebraVogel94349:
This doens't solve the problem.

PR #899, opened 2023-08-27
fix: check type of imported files

ZebraVogel94349:
Specifying allowedExtensions for a file picker does only work with specific supported file types (not .sbn and .sbn2). Because of that, the file picker didn't work on Android. Also, on platforms were the file picker worked, you could select any file, not just .sbn, .sbn2 and .pdf files. I changed it so that an error message is shown if you try to import an unsupported file and the file picker works on android now.


adil192:
Thanks!

PR #911, opened 2023-09-06
i18n: Update Spanish

myanesp:
Jay! First of all sorry for the delay, I've been on holidays and busy this last months :)

Here are the Spanish strings updated, hope everything is fine.

Keep the work up!


adil192:
Of course no worries, and thank you!

PR #924, opened 2023-09-18
i18n: Update French

Polux49:
I'm unsure about the meaning of "panning" in "Unlock panning to horizontal or vertical". I've found it in the app, but I can't see what it does.


adil192:
Panning is when you drag around on a touchscreen to move the page.
If you lock panning to the axes, you can't pan diagonally so it'll only move either horizontal or vertical

adil192:
Thanks!

Issue #170, opened 2022-12-26
Selection/move tool


Luciogi:
**Is your feature request related to a problem? Please describe.**
There is no option to move content of page

**Describe the solution you'd like**
Add option in tool bar for selection of content on page 

**Describe alternatives you've considered**
I have to remove content  and write again

**Device (please complete the following information):**
- Device: Redmi 9T
- OS: Android 12L
- Version: v0.4.9 F-Droid (409)

**Additional context**
none

**Features**
- [X] Selecting strokes
- [x] Selecting images
- [ ] Duplicate selection
- [ ] Delete selection
- [x] Change color of selected strokes

NicholasLam12:
I think it would also be useful if the selection/move tool is bundled with a copy function so that if you for some reason need two identical diagrams you wouldn't need to redraw the whole diagram again

adil192:
v0.8.0 has a new select tool. It can only move strokes around at the moment but this will be extended in future releases :)

Polux49:
+1 for the duplicate or delete. 
The following actions on the content of a selected area would also be very useful in my opinion: 

- [ ] Stretch it (suggestion: after a long press on the selected area, a box containing it would be convenient for stretching). 
- [ ] By means of the toolbar buttons, change some properties (not only color, but also width and pen style) simultaneously for all the lines in the selected area.

My guess is that you already planned it, but in case of... :-)

WasteOfO2:
You can _technically_ move ur selection to the edges and make it disappear, but i am sure this is more likely to be a bug more than a feature :P

https://user-images.githubusercontent.com/45941793/227790605-5507dd7f-a37c-4623-a1bd-2dd07072ff81.mp4

Parameter15:
Any advance on this?

KainAlive:
In other note-taking apps (GoodNotes, for example), it's possible to select a line by selecting part of it with the selection tool. In Saber, you have to select the whole stroke to move it, which is quite difficult, especially when strokes are close together. It would be nice to be able to select just part of the stroke.

adil192:
> you have to select the whole stroke to move it

@KainAlive The current behaviour is to select a stroke if most (70%) of the stroke is in the selection area. This handles cases where strokes are close together quite gracefully in my testing, such as avoiding the tail of the letter "y" in the line above when selecting something in the line below it. 

[...]

PR #926, opened 2023-09-18
feat: change color of selected strokes

AraxTheCoder:
This adds the ability to change the color of the selected strokes, as requested in #170. _canvas_painter is changed so that the user gets feedback on the changed color while keeping the selection feedback.


lib/components/canvas/_canvas_painter.dart
```diff
@@ -56,7 +56,7 @@ class CanvasPainter extends CustomPainter {
           needToRestoreCanvasLayer = true;
         }
         if (currentSelection?.strokes.contains(stroke) ?? false) {
-          paint.color = primaryColor;
+          paint.color = Color.lerp(stroke.strokeProperties.color.withInversion(invert), Colors.black, 0.25) ?? primaryColor;
```
adil192, line 59: Why did you choose `Colors.black` over `primaryColor`?

adil192, line 59: The `??` isn't necessary since neither argument is null. We can just use a `!` here and in the other line

lib/data/editor/editor_history.dart
```diff
@@ -128,6 +131,7 @@ class EditorHistoryItem {
   final Rect? offset;
   final EditorPage? page;
   final DocChange? quillChange;
+  final Map<Stroke, ColorChange>? colorChange;
```
adil192, line 134: It might be good to `assert` that the length of the `colorChange` map is the same as the list of `strokes`

lib/data/tools/select.dart
```diff
@@ -30,6 +30,33 @@ class Select extends Tool {
...
+
+    Color dominantColor = colorDistribution.keys.first;
+    int maxSize = colorDistribution.values.first;
+
+    colorDistribution.forEach((color, size) {
+      if(maxSize < size){
+        maxSize = size;
+        dominantColor = color;
+      }
+    });
```
adil192, line 55: This could be cleaner as a `reduce` call if `Map` supports it (perhaps with `map.entries`).

lib/pages/editor/editor.dart
```diff
@@ -1246,6 +1253,22 @@ class EditorState extends State<Editor> {
                 (currentTool as Highlighter).strokeProperties.color = color.withAlpha(Highlighter.alpha);
               } else if (currentTool is Pen) {
                 (currentTool as Pen).strokeProperties.color = color;
+              } else if(currentTool is Select){
```
adil192, line 1256: Please check if there's anything selected before proceeding with `select.doneSelecting`

adil192 (changes requested):
Looks good besides some small details

lib/components/canvas/_canvas_painter.dart
AraxTheCoder: I chose black so that the color is only darkened and does not appear distorted due to the primaryColor.
With the primaryColor the balance between selection feedback and color change feedback is not given.

Here are some example pictures:
https://imgur.com/a/t5IBtnO

Notice how you cant really see that "Test" was selected sucessfully.
But with the current implementation you have the same effect with black text.

adil192: > But with the current implementation you have no selection feedback on black text.

That was my concern. I think black text should take priority over coloured text since most people write in black.

Please can you change `Colors.black` to `primaryColor`?

AraxTheCoder: This is how it looks now:
Screenshot_2023-09-21-21-15-08-232_com adilhanney saber
Screenshot_2023-09-21-21-15-16-646_com adilhanney saber
Screenshot_2023-09-21-21-18-48-146_com adilhanney saber
Screenshot_2023-09-21-21-19-02-960_com adilhanney saber

lib/data/editor/editor_history.dart
```diff
@@ -115,11 +116,13 @@ class EditorHistoryItem {
...
     this.page,
     this.quillChange,
+    this.colorChange,
   })  : assert(type != EditorHistoryItemType.move || offset != null, 'Offset must be provided for move'),
         assert(type != EditorHistoryItemType.deletePage || page != null, 'Page must be provided for deletePage'),
         assert(type != EditorHistoryItemType.insertPage || page != null, 'Page must be provided for insertPage'),
         assert(type != EditorHistoryItemType.quillChange || quillChange != null, 'Quill change must be provided for quillChange'),
-        assert(type != EditorHistoryItemType.quillUndoneChange || quillChange != null, 'Quill change must be provided for quillUndoneChange');
+        assert(type != EditorHistoryItemType.quillUndoneChange || quillChange != null, 'Quill change must be provided for quillUndoneChange'),
+        assert(type != EditorHistoryItemType.changeColor || colorChange != null || colorChange!.length == strokes.length, 'ColorChange must be provided and must have the same length as strokes for changeColor');
```
adil192, line 125: I don't think this binary logic would work

```suggestion
        assert(type != EditorHistoryItemType.changeColor || colorChange?.length == strokes.length, 'ColorChange must be provided and must have the same length as strokes for changeColor');
```

lib/data/tools/select.dart
```diff
@@ -30,6 +30,27 @@ class Select extends Tool {
     selectResult.pageIndex = -1;
   }
 
+  Color? getDominantStrokeColor(){
```
adil192, line 33: For completeness we could add a `doneSelecting` check here. What do you think?
```suggestion
  Color? getDominantStrokeColor(){
    if (!doneSelecting) return null;

```

adil192 (changes requested):

adil192:
Tested working as expected, thanks :)

lib/data/tools/select.dart
AraxTheCoder: Makes sense.
Didn't had the whole ``doneSelecting`` thing in mind.
Definitely gonna look twice for future pull requests.

adil192 (dismissed):

adil192:
Thanks!

Issue #922, opened 2023-09-16
Bulk Exports / Auto Exports to PDF / Save notes to public storage on Android


andrigamerita:
### Motivation

It's good that the Android app includes both a Nextcloud sync feature, and a manual export of single notes, but they're bot pretty limited solutions in practice, at least for me.

One of my desires is to be able to add some handwritten notes to, for example, my website, but making the update process when I edit a note as automatic and transparent as possible. This means that I should have a way to export all notes I care about in an instant, either on command for multiple notes, or automatically after setting up the feature.

I think my ideas can be useful not only for people who share their notes, but also other ones that want to manage them via external "cloud" solutions, like Git. Especially considering that the Nextcloud support in the app has strict server requirements.

### Solution

The first great thing would be to have a multi-selection option in the folders/notes list, this way one could select all interested notes and export them to the local file system, in either SBN2 or PDF format. (Currently, only one note at a time can be exported, only by opening it and clicking share.)

Preferably, the app should also let the user store all notes in the Android public storage instead of `/data/data/...`, which is only readable by other processes with root, and thus impossible to get files out externally for legitimate purposes on unmodded devices. By providing this option, scripts and other apps can get access to the Saber notes without roadblocks. (At the same time, I recognize that saving to public storage could mean the notes aren't safe from tampering/reading by malicious apps, and so yes, this should be an optional feature, not an hard change.)

Finally, if the user is choosing to save notes to the public storage, maybe there should be an option to automatically also save each note always in PDF format, in addition to SBN2, so it's possible to immediately share these on other places. In my case, I could launch a script that with just one click uploads all these PDFs, be they old or new, to my website.

All the things I've mentioned above could be implemented one by one, independent from each other, and all would make for great improvements with little effort, in my opinion. As far I know, if the app writes to it's own private directory in public storage (`/sdcard/Android/data/...`), it won't even have to ask for user permission and implement the complex Android file-system access APIs, so it really shouldn't be hard to make the app write to the public storage in this fashion.

### Alternatives

My only fear is that bulk PDF save/export, or automatic export on save, could be slow and clunky, since I've seen that the app takes quite a while to generate a PDF file when clicking share for a single note. My guess is that's an expensive operation on the compute side, but I hope I'm mistaken and instead the app would be able to manage such a thing.

[...]

PR #940, opened 2023-10-03
feat: select multiple notes to move/rename/delete

ZebraVogel94349:
This replaces the buttons that appear when you select a note by a bar on the bottom which these buttons on it. When multiple files are selected at once, it is now possible to move or delete them (exporting/sharing multiple notes at once, like suggested in #922 could be added to the bar as well). Renaming is only possible, when only one note is selected. Otherwise, the button is hidden. Also, this fixes a bug that caused the selection of a note to disappear when the note is not visible anymore after scrolling.

### DIFF
--- a/lib/components/home/move_note_button.dart
+++ b/lib/components/home/move_note_button.dart
@@ -138,7 +138,7 @@ class _MoveNoteDialogState extends State<_MoveNoteDialog> {
       parentFolderCounts[parentFolder] = (parentFolderCounts[parentFolder] ?? 0) + 1;
     }
     return parentFolderCounts.entries
-        .reduce((a, b) => a.value > b.value ? a : b)
+        .reduce((a, b) => a.value >= b.value ? a : b)
         .key;
   }
 
--- a/lib/components/home/move_note_button.dart
+++ b/lib/components/home/move_note_button.dart
@@ -70,14 +70,23 @@ class _MoveNoteDialogState extends State<_MoveNoteDialog> {
   /// destination folder. In that case, the file name
   /// will be suffixed with a number.
   List<String> newFileNames = [];
+  List<String> duplicateFileNames = [];
 
 
   Future findChildrenOfCurrentFolder() async {
     currentFolderChildren = await FileManager.getChildrenOfDirectory(currentFolder);
+    newFileNames = [];
+    duplicateFileNames = [];
     for(var i = 0; i < widget.filesToMove.length; i++){
-      newFileNames.add(await FileManager.suffixFilePathToMakeItUnique('$currentFolder${widget.filesToMove[i]}', false, '${parentFolders[i]}${widget.filesToMove[i]}${Editor.extension}')
-      .then((newPath) => newPath.substring(newPath.lastIndexOf('/') + 1)));
+      String newFileName = await FileManager.suffixFilePathToMakeItUnique('$currentFolder${widget.filesToMove[i]}', false, '${parentFolders[i]}${widget.filesToMove[i]}${Editor.extension}')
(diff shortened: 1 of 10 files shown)


lib/components/home/move_note_button.dart
```diff
@@ -38,49 +36,46 @@ class _MoveNoteDialog extends StatefulWidget {
...
+  late List<String> fileNames = widget.filesToMove.map((path) => path.substring(path.lastIndexOf('/') + 1)).toList();
+
+  /// The original parent folder of the notes,
   /// including the trailing slash.
-  late String parentFolder = widget.existingPath.substring(
-    0,
-    widget.existingPath.lastIndexOf('/') + 1,
-  );
+  
+  late String parentFolder = widget.filesToMove[0].substring(0, widget.filesToMove[0].lastIndexOf('/') + 1);
```
adil192, line 54: I don't think the `filesToMove` always have the same parent folder. Maybe this would be better as a `List<String>` like `fileNames`?

lib/components/home/move_note_button.dart
```diff
@@ -153,8 +148,8 @@ class _MoveNoteDialogState extends State<_MoveNoteDialog> {
                 ],
               ),
             ),
-            if (newFileName != null && newFileName != fileName)
-              Text(t.home.moveNote.renamedTo(newName: newFileName ?? '?')),
+            if (newFileNames != null && newFileNames!.length == 1 && newFileNames![0] != fileNames[0])
+              Text(t.home.moveNote.renamedTo(newName: newFileNames![0])),
```
adil192, line 152: I think this could be improved to support multiple files, something like this:
```dart
if (newFileNames != null)
  for (int i = 0; i < fileNames.length; ++i)
    if (fileNames[i] != newFileNames![i])
      Text(t.home.moveNote.renamedTo(newName: newFileNames![i])),
```

But we could change the text to be like this if we have more than one changed file name (for only one changed file name, the old one is better)
```
Some files will be renamed when moved into this folder:
- somefile (2)
- notes (3)
```

lib/components/home/preview_card.dart
```diff
@@ -197,8 +201,8 @@ class _PreviewCardState extends State<PreviewCard> {
     Widget card = MouseRegion(
```
adil192, line 201: I think we could change this so that if we have some files already selected, tapping another (unselected) file will add it to the selection rather than opening the note.
(For this we'd need to conditionally call the `action` parameter from below: `closedBuilder: (context, action) => card`.)

lib/pages/home/recent_notes.dart
```diff
@@ -121,6 +129,49 @@ class _RecentPageState extends State<RecentPage> {
...
+                child: RenameNoteButton(existingPath: selectedFiles.length == 1 ? selectedFiles[0] : '')
+              );
+            },
+          ),
+          ValueListenableBuilder(
+            valueListenable: selectedFiles,
+            builder: (context, selectedFiles, child) {
+              return Collapsible(
+                axis: CollapsibleAxis.vertical,
+                collapsed: selectedFiles.length != 1,
```
adil192, line 149: You've updated the `move` function but hidden the move button if there's multiple files.
Was this intentional?

adil192 (changes requested):

lib/pages/home/recent_notes.dart
ZebraVogel94349: This was to ensure that the parent folder is always the same. Now different parent folders are possible, so I changed it to be like in the "Browse" screen.

adil192:
Thanks!

PR #953, opened 2023-10-09
i18n: Update Czech

ceskyDJ:


ceskyDJ:
> Also note that editor.selectionBar.delete was copied from editor.imageOptions.delete so isn't marked as (OUTDATED), but feel free to edit it if needed.

Checked, should be ok. So, feel free to merge it ;).

adil192:
Thanks!

PR #956, opened 2023-10-14
i18n: Update German

hubortje:
Minor changes (two words) to be more precise and consistent with the rest of the translation.

„Löschen Sie $f“ uses the obsolete formal you, „$f löschen“ is more precise and in "tune" with the rest of the strings.

The change from "wenn" to "sobald" is more subtle: "wenn", in this context, works like an "if", but it is not an if-case and rather an "as soon as"-case aka. "sobald".


adil192:
Thanks!

PR #958, opened 2023-10-16
i18n: Add Hebrew

Elielv235:
made an hebrew(israel) translation


adil192 (changes requested):
Thank you!
I've noticed that in your PR and commits, you sometimes use locale code `he` and sometimes use `he_IL`.
`he` would be preferable if there's no significant differences between `he_IL` and other Hebrew locales, like how I use "English" instead of "UK English".
Of course, you're the expert here on Hebrew, so which would you prefer?

adil192:
I'm going to go with just `he` because there are no other sub-locales listed on [the link from the README](https://saimana.com/list-of-country-locale-code/).

adil192 (approved):

adil192:

Elielv235:
yes i do! and there is no difference between he_IL and he cuz IL its just the country code

Elielv235:
and when i will can apply Hebrew into the app itself(because the option isn't showing up)

adil192:
> and when i will can apply Hebrew into the app itself(because the option isn't showing up)

Are you on v0.17.0?

PR #963, opened 2023-10-22
fix: update multi-selection buttons after changing selection

ZebraVogel94349:
Currently, the button to rename a note is not hidden when multiple notes are selected. This is because selectedFiles just contains a reference to the values stored in the list, which doesn't change when you change the content of the list (except for when it's empty and you add something or when you remove everything from the list), so the ValueListenables are not called. By creating a new list every time the selection changes, the reference changes and therefore the rename button is hidden when selecting two notes or more. You can also call selectedFiles.notifyListeners(), but that causes some warnings.


adil192:
Since the list object itself isn't changing, the `setSelectedFiles` is a bit meaningless. I've swapped it out for just calling notifyListeners

PR #972, opened 2023-10-25
feat: export notes from home screen

ZebraVogel94349:
This adds a button export/share all selected notes in the "Browse" and "Recent Notes" tabs on the home screen, like suggested in #922 . When you press the button a menu like the one for creating notes opens where you can choose whether to export it as SBN2 or PDF. When multiple notes are selected, a zip file with all of the exported notes gets created.


adil192:
Sorry for the delay!

PR #977, opened 2023-10-27
fix: import sbn/sbn2 when in root folder

ZebraVogel94349:
This fixes the import of .sbn and .sbn2 files into the root directory. When widget.path was null, importFile() was called with the parameter 'null/' which caused the file not to be imported. Also, this makes the imported note open correctly after importing by removing the file extension from the path (previously, an empty .sbn2.sbn2 file was opened).

### DIFF
--- a/lib/components/home/move_note_button.dart
+++ b/lib/components/home/move_note_button.dart
@@ -126,7 +126,7 @@ class _MoveNoteDialogState extends State<_MoveNoteDialog> {
   @override
   void initState() {
     currentFolder = _findMostCommonParentFolder();
-    if(!currentFolder.startsWith('/')) {
+    if (!currentFolder.startsWith('/')) {
       currentFolder = '/$currentFolder';
     }
     super.initState();
--- a/lib/components/home/move_note_button.dart
+++ b/lib/components/home/move_note_button.dart
@@ -126,6 +126,9 @@ class _MoveNoteDialogState extends State<_MoveNoteDialog> {
   @override
   void initState() {
     currentFolder = _findMostCommonParentFolder();
+    if(!currentFolder.startsWith('/')) {
+      currentFolder = '/$currentFolder';
+    }
     super.initState();
 
     findOldExtensions()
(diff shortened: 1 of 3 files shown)


lib/components/home/move_note_button.dart
```diff
@@ -126,6 +126,9 @@ class _MoveNoteDialogState extends State<_MoveNoteDialog> {
   @override
   void initState() {
     currentFolder = _findMostCommonParentFolder();
+    if (!currentFolder.startsWith('/')) {
+      currentFolder = '/$currentFolder';
+    }
```
adil192, line 131: If the old imported files start with `null/...` then shouldn't we check for that and remove the `null` prefix?

ZebraVogel94349: The files don't start with null. They are in a Folder called Sabernull. And if you tried move them using the old version, they were moved into other folders starting with "Saber", because the / at the beginning of the path was missing. So you have to check for all folders in the same parent directory as the Saber folder that start with "Saber".

adil192: Wouldn't the only ones that get to this point be recent files, which are already in the form `null/...`?

ZebraVogel94349: > Wouldn't the only ones that get to this point be recent files, which are already in the form `null/...`?

What exactly do you mean by that? The files are not in the form `null/...`, they just go imported into the folder `Sabernull` (`/home/user/Sabernull` on Linux) instead of `Saber`. The path after that and the filename were correct.

adil192: If files are imported to `null/...` then they will be written on the file system to `Documents/Sabernull/...` but in Saber, they will be stored in `Prefs.recentFiles` in the original `null/...` form

ZebraVogel94349: I fixed it by moving all files in Prefs.recentFiles that don't start with a `/` to the same path but with an added `/` in the beginning or with the `null` prefix removed, if they start with `null`.

adil192:
Thanks!

PR #985, opened 2023-10-30
i18n: Update Turkish

fybx:


fybx:
I've resolved the merge conflict caused by `e0f4652` and others while translating new strings 🥇

PR #1023, opened 2023-11-30
i18n: Update Spanish

myanesp:
Hi again!

New strings translated :) Now I have more free time so I will try to keep this updated as possible.

Thanks!

### DIFF
--- a/lib/i18n/_missing_translations.yaml
+++ b/lib/i18n/_missing_translations.yaml
@@ -141,34 +141,19 @@ de:
       bgPatterns:
         tablature(OUTDATED): Tablature
 es:
-  home:
-    tooltips:
-      exportNote(OUTDATED): Exportar nota
-    moveNote:
-      moveNotes(OUTDATED): Mover $n notas
-      multipleRenamedTo(OUTDATED): "Renombrar las siguientes notas"
-      numberRenamedTo(OUTDATED): $n notas serán renombradas para evitar conflictos
   settings:
     prefCategories:
       performance(OUTDATED): Performance
     prefLabels:
-      recentColorsLength(OUTDATED): Número de colores recientes a almacenar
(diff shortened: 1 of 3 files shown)


lib/i18n/_missing_translations.yaml
```diff
@@ -137,27 +137,27 @@ de:
...
   settings:
     prefLabels:
-      recentColorsLength(OUTDATED): How many recent colors to store
+      recentColorsLength(OUTDATED): Número de colores recientes a almacenar
   editor:
     pens:
-      pencil(OUTDATED): Pencil
-      shapePen(OUTDATED): Shape pen
+      pencil(OUTDATED): Bolígrafo
+      shapePen(OUTDATED): Bolígrafo con formas
```
adil192, line 151: The Spanish translation already has `ballpointPen: Bolígrafo`. Are you able to reword pencil, like to Lápiz?

lib/i18n/_missing_translations.yaml
```diff
@@ -137,27 +137,27 @@ de:
...
     colors:
-      customBrightnessHue(OUTDATED): Custom $b $h
-      customHue(OUTDATED): Custom $h
-      dark(OUTDATED): dark
-      light(OUTDATED): light
+      customBrightnessHue(OUTDATED): $b $h personalizado
```
adil192, line 153: If I understand correctly, Spanish puts its adjectives after the noun. An example `customBrightnessHue` in English would be `Custom dark blue`: would this be `Azul Oscuro personalizado` ($h $b) or `Oscuro Azul personalizado` ($b $h)?

adil192 (changes requested):

adil192:
I made those changes for now, but left them as OUTDATED so you can edit them in the future if need be

Issue #1019, opened 2023-11-27
Line detection


Iey4iej3:
### Motivation

This functionality was removed when the shape tool was introduced. However, there are several advantages of line detection over the shape tool:

1. As mentioned in https://github.com/saber-notes/saber/discussions/970#discussion-5771538 the lines were perfectly horizontal or vertical, while the current shape tool does not produce such perfect lines.
2. The shape tool has a much heavier computational overhead. The lag is obvious with stylus, and it is quite intolerable on, say, Onyx Boox.

### Solution

Bring back the line detection functionality.

### Saber version

v0.17.2 F-Droid (17020)

### Device

- Device: -
- OS: -

adil192:
I'm working on a faster shape recognition package, see [one_dollar_unistroke_recognizer](https://pub.dev/packages/one_dollar_unistroke_recognizer).
I'll add back line snapping when it's ready, as well as line recognition without needing to use the shape pen.

adil192:
@Iey4iej3 How is the lag now with v0.18.3?

Iey4iej3:
I will test it on my Onyx Boox when it is available on F-Droid. However, I still think that the previous line detection is of use (although it seems easier than before to produce perfect horizontal or vertical lines via the shape pen, as I tested on my laptop in the flatpak version). It can be done without switching the mode, which seems useful when I am taking notes.

Iey4iej3:
The line detection is back in v0.21.3.

PR #1025, opened 2023-12-02
feat: setting to show/hide shape pen preview

ZebraVogel94349:
The lag of the shape pen described in #1019 is mostly caused by rendering the shape preview 10 times per second. This adds a setting to hide the preview, which greatly reduces the lag.


Iey4iej3:
The "corrected" shape would be unpredictable if the preview is disabled.

adil192:
I think this would be better as a multiple choice (e.g. Fast for 0.1s, Slow for 0.5s, Disabled for no previews).
I'm going to close this PR to reimplement it with these choices, and put it in a Performance section in settings

PR #1049, opened 2023-12-10
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Hi @adil192! This maybe sound very easily, but I thing it will be very hard to understand to general users. In English it maybe isn't that bad, but different languages like Czech has problems with that. I tried to translate it as good as I know, but I don't know it's good enough.

Where is the button located? I tried to find it, but I wasn't successful. Maybe some additional text could explain it better than long text inside button. If there could be an explanation text close to the button, text inside the button could be just "Synchronize" or "Resynchronize", which is pretty well translatable and understandable).

What do you think?

adil192:
The button is here by the login status on the Settings page:
image


> text inside the button could be just "Synchronize" or "Resynchronize"

That sounds fine :) I don't expect that people use this button very much anyway but "Resynchronise" is perfect

ceskyDJ:
Oh, I see. So, I'll add it more detail comment, it should fit there.

ceskyDJ:
So, I finally ended on something very similar to my original proposal. I think it could be well-readable and understandable. So, feel free to merge it ;).

adil192:
Thanks!

PR #1053, opened 2023-12-13
github: split up tests between runners

adil192:


adil192:
There was no speed benefit at all :(
image

PR #1058, opened 2023-12-16
fix(browse): deselect after deleting selected files

ZebraVogel94349:
This fixes the deselection of files after deleting them in the "Browse" tab.  571d571 only fixes it in the "Recent notes" tab.


adil192:
Thanks!

PR #1061, opened 2023-12-17
fix: unselect notes after moving

ZebraVogel94349:
This fixes unselecting notes after moving them. Also this fixes a bug which caused the first click on a note after deleting selected notes to result in the note being selected instead of opened in the editor.


adil192:
Thanks!

Issue #1080, opened 2023-12-30
Nextcloud sync - Resync everything buton causes multiple uploads at one time


QubaB:
### Describe the bug

I want to synchronise all my notes to nextcloud using "Resync everything". I see big traffic on my own nextcloud server.

See also my discussion https://github.com/saber-notes/saber/discussions/1074

### To reproduce

1. Create a lot of notes
2. Click on Resync everything button
3. On nextcloud server is large number of upload threads

### Expected behavior

Only one note at a time is uploaded

### Saber version

v0.19.0

### Device

- Device: Samsung Galaxy Tab S6
- OS: Android 12

QubaB:
Fixed in v0.19.3

PR #1081, opened 2023-12-30
Nextcloud sync - Resync everything buton causes multiple uploads at one time

QubaB:
"Resync everything" button causes multiple uploads at one time. This commit avoids it using multiple tests of _isUploadingFile.  Solves issue #1080


QubaB:
This pull request is also part of request #1083. I supposed they will be distinguished but are not.

Issue #1082, opened 2023-12-30
Resync everything button cause upload of zero size files to nextcloud


QubaB:
### Describe the bug

Press of Resync everything button cause upload of zero size files to nextcloud server.

See my investigation in https://github.com/saber-notes/saber/discussions/1074

### To reproduce

1. Create some notes
2. Press Resync everything button

### Expected behavior

Resync everything should upload all notes files to nextcloud server

### Saber version

v0.19.0

### Device

- Device: Samsung Galaxy Tab S6
- OS: Android 12

PR #1083, opened 2023-12-30
Resync everything button cause upload of zero size files to nextcloud

QubaB:
Pressing "Resync everything" button causes upload of zero size files to nextcloud because getChildrenOfDirectory removes extensions from file names. Saber is uploading files without extension then. They do not exist so zero size files are created.

This commit fixes behavior of getChildrenOfDirectory  adding optional parameter removeExtension (by default set to true for compatibility). Also getAllFiles is modified to propagate parameter.

Fixes issue #1082  and #1081


lib/data/file_manager/file_manager.dart
```diff
@@ -390,18 +393,19 @@ class FileManager {
...
+            // remove extension
+            if (filePath.endsWith(Editor.extension)) {
+              filePath = filePath.substring(
+                  0, filePath.length - Editor.extension.length);
+            } else if (filePath.endsWith(Editor.extensionOldJson)) {
+              filePath = filePath.substring(
+                  0, filePath.length - Editor.extensionOldJson.length);
+            }
+
+            if (Editor.isReservedPath(filePath))
```
adil192, line 406: I see that you've moved this into the `if (removeExtension)` block.
I think it'd make more sense if we add another parameter for `includeReservedPaths` rather than tying it into an unrelated flag

lib/data/file_manager/file_manager.dart
```diff
@@ -372,7 +372,10 @@ class FileManager {
   }
 
   static Future<DirectoryChildren?> getChildrenOfDirectory(
-      String directory) async {
+      String directory,
+      [bool removeExtension = true]) async {
```
adil192, line 376:
```suggestion
      {required bool removeExtension}) async {
```
I'd prefer a named parameter in this case as otherwise it's hard to read (will need some refactoring of course)

lib/data/nextcloud/file_syncer.dart
```diff
@@ -161,17 +161,45 @@ abstract class FileSyncer {
   static Future uploadFileFromQueue() async {
     if (_isUploadingFile) return;
     await _uploadQueue.waitUntilLoaded();
-    if (_uploadQueue.value.isEmpty) return;
-
+    if (_uploadQueue.value.isEmpty) {
+      return;
+    }
+    if (_isUploadingFile){
```
adil192, line 167: These `FileSyncer` changes look like duplicates of #1081. If nothing new is here (?), I'll just disregard it when I rebase this branch.
(Also please avoid updating the branch via merge commits as it makes the commit history harder to read)

lib/components/settings/nextcloud_profile.dart
```diff
@@ -116,10 +116,25 @@ class _NextcloudProfileState extends State<NextcloudProfile> {
                   ),
                   tooltip: t.settings.resyncEverything,
                   onPressed: () async {
-                    final allFiles = await FileManager.getAllFiles();
+                    final allFiles = await FileManager.getAllFiles(false);
                     Prefs.fileSyncResyncEverythingDate.value = DateTime.now();
                     for (final file in allFiles) {
                       FileSyncer.addToUploadQueue(file);
+                      // upload also all file assets if any
```
adil192, line 123: Rather than using this for-loop which is quite slow, I think we should add a `{bool includeAssetFiles = false}` parameter into `FileManager.getChildrenOfDirectory`

adil192 (changes requested):

lib/data/file_manager/file_manager.dart
QubaB: I suggested some code, which worked for me. I am not familiar with programming dart and flutter applications so my code is not optimal (I am FORTRAN programmer of scientific computations). 

Please modify the code completely according to your ideas. I don't stand by what I wrote at all. I'm not used to collaborating on code creation, so I would ask you to process everything yourself.

lib/data/nextcloud/file_syncer.dart
QubaB: Yes, I forgot to create new branch. I see that you already closed #1081.

lib/components/settings/nextcloud_profile.dart
QubaB: Yes, it seems to be good idea.

lib/data/file_manager/file_manager.dart
QubaB: I will try to apply your suggestions and commit it. Hope it saves you some time

lib/data/file_manager/file_manager.dart
```diff
@@ -390,18 +396,36 @@ class FileManager {
...
-          // remove extension
-          if (filePath.endsWith(Editor.extension)) {
-            filePath = filePath.substring(
-                0, filePath.length - Editor.extension.length);
-          } else if (filePath.endsWith(Editor.extensionOldJson)) {
-            filePath = filePath.substring(
-                0, filePath.length - Editor.extensionOldJson.length);
-          }
-
+          //log.info('Listing file $filePath');
```
droghio, line 399: nit: Seems like dead code, may with to remove.

lib/data/file_manager/file_manager.dart
```diff
@@ -390,18 +396,36 @@ class FileManager {
...
+            } else if (filePath.endsWith(Editor.extensionOldJson)) {
+              filePath = filePath.substring(
+                  0, filePath.length - Editor.extensionOldJson.length);
+            }
+            else {
+              return null; // filePath is name of some asset
+            }
+          }
+          else {
+            if (includeAssetFiles){
```
droghio, line 417: nit: I think you can remove a layer of if nesting.

Disclaimer: I'm not involved in this project but happened along the PR when trying to debug the same issue. So thanks for making a fix! Noticed a couple nits but again I'm just a random SWE so feel free to ignore.

I could be off here but IIUC this block is primary meant to return early to avoid including asset files when the argument is false If so it may be simpler to pull this if check into the inner most else down below and remove a layer of nesting.

We may be also be able to flip the if check to return the path if valid vs. return early if not. Given the arguments here it may simplify things.

I.e. something like:

```
if (includeAssetFiles || (is old note) || (is new note)){ return ... add file }
else { return null }
```

The main advantages being a little easier to read conditional (although that is a bit of an opinion on my part) and reduced nesting.

lib/data/file_manager/file_manager.dart
QubaB: If removeExtension is true, then we should return only file names ending sbn2 or sbn. So it should be there I think

droghio: Sorry I mean the log statement that is commented out.

lib/data/file_manager/file_manager.dart
QubaB: If removeExtension is false, then there are two possibilities: 
1. includeAssetFiles is true, then every file name should be returned. 
2. if includeAssetFiles is false, then only file names sbn and sbn2 should be returned.

In fact are now used only two combinations:

removeExtension=true and includeAssetFiles=false (original code which lists only names of notes)
removeExtension=false and includeAssetFiles=true (code used in case of resync all which returns every file name)

So the code for removeExtension=true and includeAssetFiles=true  which are you discussing will never be interpreted and it will not be slow.

I think that it is OK and I would leave the decision on @adil192

QubaB (commented):
I finished review of code

adil192 (approved):

adil192:
Sorry for the delay, I've been quite busy recently. Thanks for the fix!

adil192:
Before I merge, is `Jakub Zlámal <zlamaljakub@gmail.com>` your email?
If so, I'll disregard it as a co-author in favor of @QubaB.

QubaB:
Yes, zlamaljakub@gmail.com is my email.   

Thank you for accepting (and improving) this fix.

Issue #1084, opened 2023-12-30
Better language signpost in `README.md`


ceskyDJ:
Currently we have a signpost for various languages of "README" in `README.md`, that looks like this:
![image](https://github.com/saber-notes/saber/assets/48548230/c78aa281-68ad-4af3-98eb-e5395d82b594)

It has some problems in languages like Czech, when there is more complex rules for size of the first letter of language name (Čeština vs čeština), because there isn't always big letter like in other languages (like e.g. English and German). The second problem is word inflection. The sentence "You can read this file in" requires the language name to be in a special form, so it's not consistent and I'm sure it's not too clear for understand.

I propose changing our signpost to simpler form used by tuyafeng/Via:
![image](https://github.com/saber-notes/saber/assets/48548230/a83c7287-8a7a-4b0b-bfa1-86ee13a23f55)

Here are just language names divided by a pipe. Language names could be in a base form and with big first letter. It's simple and well understandable, at least I think. Of course, logo position isn't subject of this discussion, Saber has just perfect position of it. What do you mean?

AkintolaRichard:
@ceskyDJ this is awesome and I would love to join in implementing this.
@adil192 please can you can provide guidance implementing this?

ernest-emmanuel-utibe:
@ceskyDJ great job
@adil192 Like @AkintolaRichard has said, please provide enough guidance for implementing this.

To me, the proposed solution suggests simplifying the format to just language names divided by a pipe, using a base form of the language name with an initial capital letter.

Instead of using full-language names, consider utilizing standardized language codes like ISO 639-1 codes (two-letter codes) or ISO 639-3 codes (three-letter codes). This could provide a more concise and universally recognized way to represent languages.

If space allows or in a dynamic setting (like a web interface), consider implementing a tooltip or hover feature that displays the full language name when users hover over or click on the language code or name. This can provide additional clarity without cluttering the primary signpost

This simpler format aims to avoid issues related to complex capitalization rules and word inflection that might occur in certain languages.

ceskyDJ:
Hi @AkintolaRichard and @ernest-emmanuel-utibe! I don't completely understand, what do you want from me. So, could you be more specific, please?

[...]

PR #1093, opened 2024-01-04
Better language signpost in README.md #1084

Viswesh934:
In this version, each language link is accompanied by a brief description of what users can expect when they click on the link. Adjust the descriptions as needed to accurately represent the content available in each language

image

Modified capitalization and adjusted word inflation


ceskyDJ:
I think you don't read #1083 carefully. It's just a discussion about possible change. Here are more questions about translating in Saber waiting for @adil192's choice. I think it should be done at once after a successful discussion with the repository owner.

Anyway, if I look at your PR as a proposal, it doesn't solve issues noted in #1083. It also makes the signpost even more complicated. I don't know, if this is the way we want to select.

ceskyDJ:
As I see in your profile you're from India. We don't have Saber localized for India. Could you do it, please? I think, this will be better to delay a bit and discuss a bit about it. You can translate README and the Saber alone to your native language, it will help a lot. After that, I hope you will be maintaining the translation since Saber is in active development and changes still come.

PR #1094, opened 2024-01-04
Update README languages toggling

shemigam1:


ceskyDJ:
Hi, I don't know what it is but it is at least very low quality change. I think you don't know, what the changes are about. I'm sorry for these words, but this can't be merged.

There was an issue about this change, but it's just a proposal with discussion. It waits for @adil192's opinion.

ceskyDJ:
Why did you create this, when you can see, there is already a very similar PR #1093?

adil192:
Closing this #1084 as there are mistakes and it's a very small change that doesn't need a PR

PR #1100, opened 2024-01-13
feat: confirmation prompt for deleting selected notes

ZebraVogel94349:
This adds a confirmation prompt for deleting selected notes, so you cannot accidentally delete notes by miss clicking if you want to export or move them. The prompt works exactly like the one for deleting folders.


ZebraVogel94349:
I've updated this pull request to pass all tests and to make it possible to merge without conflicts again.

adil192 (approved):
Works well, thank you

PR #1107, opened 2024-01-17
i18n: Update Czech

ceskyDJ:


ceskyDJ:
I'm going to update Czech version of README, too, but I have exams and it's more time consuming task, so it must wait a bit.

adil192:
Thanks!

PR #1112, opened 2024-01-22
perf: avoid `MediaQuery.of`

ellet0:
For example, instead of `MediaQuery.of(context).size` now we will use `MediaQuery.sizeOf(context)`

Which will listen for only size changes to reduce the rebuilds and help improving the performance


ellet0:
@adil192 This is pretty easy PR to merge or reject, it doesn't change much, it could help with performance a little

adil192:
Thanks!

PR #1128, opened 2024-02-11
fix: use filepaths as cache keys, not File objects

### LINKED ISSUE #1066: Losing pdf background
### Describe the bug

Hi,
 I am using the pdf import option in order to annotate them. I also add some blank page into the pdf. These files are not huge. Usually les than 5  pages. 
Very often when I rearrange all these pages, I lose the background pdf on some of them.
I can se the circle in the middle of thr page (like buffering).
Also, I see the same thing when I click on options. 
See attached.

IMG_20231217_220932
IMG_20231217_220941

I am not able to reproduce it every time.

### To reproduce

1. Importet pdf file
2. Write on it
3. Add some blank pages to it
4. Rearange pages 
5. Remove the last blank page (if the delete button is grey, move the page ip and then remove it.)
6. Save the file 

### Expected behavior

The background of the pdf file should not disappear. 

### Saber version

0.18.4 (F-Droid)

### Device

- Device: Lenovo P12
- OS : Android 13

### Anything else?

_No response_

QubaB:
This commit fixes handling of Assets caches

1. AssetCache - key is now set to File.path (String) because File was incorrectly checked for presence in AssetCache
2. OrderedAssetCache - in add function is sometimes set index of (pdf) asset to -1 even if it is the same as already stored. So additional check is implemented to return asset.
3. Added bool EnabledRemove to AssetCache. It is used to block removing of asset during note is saved.

Closes #1066  (I hope)

### DIFF
--- a/lib/components/canvas/_asset_cache.dart
+++ b/lib/components/canvas/_asset_cache.dart
@@ -55,13 +55,18 @@ class AssetCache {
   /// Returns whether the image was present in the cache.
   bool removeImage(EditorImage image) {
     if (!allowRemovingAssets) return false;
-    for (final file in _images.keys) {
-      if (_images[file]!.remove(image)) {
-        _images.remove(file);
-        _cache.remove(file);
+
+    for (final filePath in _images.keys) {
+      final imagesUsingFile = _images[filePath]!;
+      imagesUsingFile.remove(image);
+
+      if (imagesUsingFile.isEmpty) {
+        _images.remove(filePath);
+        _cache.remove(filePath);
         return true;
       }
     }
+
     return false;
   }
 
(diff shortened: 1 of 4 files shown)


QubaB:
I think it would be useful to add bool IsBackground to PdfEditorImage.

In the current implementation, when moving to another page, PdfEditorImageloadOut() is called fro background pdf, which removes the pdf from the AssetCache, and then it is added there again at request from another page. If IsBackground was set then loadOut() would not be called.

I'm not sure how it would be handled if pages with this background were deleted until there were none left. It would probably be necessary to add a counter, on how many pages the background pdf is used, and if it is no longer used, it would be deleted.

This is too complex for me to guess implementation.

adil192:
Thanks for the heads up, looks like my logic in `AssetCache.removeImage` was wrong. It now checks whether any other `EditorImage`s are using the file before removing it from cache (like the dartdoc previously claimed).
Also since pdfs are basically always backgrounds, I think an `isBackground` flag would be similar to just never unloading pdfs.

PR #1129, opened 2024-02-12
fix: delete preview when deleting note

QubaB:
Delete also preview file  *.sbn2.p when deleting note with assets


adil192:
Thank you!

Issue #1125, opened 2024-02-09
Can't export to pdf


ThmsDhn:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

Hello

I'm using your great app with a Nextcloud server: 
- on my linux laptop (version v.0.19.2 (19020)); 
- on a Lenovo Tab P12 and on my smartphone (version v.0.19.2 F-Droid (19020)).

On each device, i have the same problem : i manage to export my projects to sba format but i can't export as pdf. The wheel is going around but nothing happens...

Could you please help me ?

### To reproduce

Export project as pdf.

### Expected behavior

Export as pdf to my Nextcloud server.

### Saber version

v.0.19.2 (19020)

### Device

Laptop :  Linux Mint LMDE 6

Tablet : Lenovo Tab P12

QubaB:
Did you tried to update to 0.19.4? There is fix of export on iOS, may be it helps.

I think that export is done only to device. Not to Nextcloud.

ThmsDhn:
Hello 
Thanks for the help !
Yes, i'm already with 0.19.4 on each device.
A friend had the same problem on iOS and 0.19.4 effectively fixed it.
Two or three months ago, i was able to export as pdf directly to my Nextcloud (or to device, or to email...).

saihaze:
Same problem, 0.20.1 on Android.

iniyanijoseph:
Same issue, it doesn't work for 20.1 on Android, on a Macbook, or on Linux (Debian Appimage).

arpandaze:
Can you guys please confirm whether this fixes for you or not: https://github.com/saber-notes/saber/pull/1135 ?

iniyanijoseph:
> Can you guys please confirm whether this fixes for you or not: #1135 ?

That was it for me. Hopefully this gets merged soon.

QubaB:
I tested two files of @ThmsDhn (Weierstrass and Riemann) and pdf was exported correctly. 

I think that @arpandaze fix works well.

Issue #1026, opened 2023-12-02
PDF exports not working


RamsaKd:
### Describe the bug

When I try to export a note as a pdf it just spins endlessly and never actually exports.

### To reproduce

1. Make a note
2. Try to export it as a pdf

### Expected behavior

It should take some time and then export

### Saber version

0.18.3 Foss android

### Device

- Device: Google Pixel 7 Pro
- OS: Android 14

adil192:
I can't reproduce this, please can you tell me what features are in your note like pdfs, images, shapes, etc?

fritzrehde:
I also experience this. I'm on the Samsung Tab S6 Lite running LineageOS 20 (Android 13). When I click on the export as PDF button, the app freezes and eventually crashes. The note in question is a long PDF (around 80 pages) where I annotated the first page with my S Pen.

fritzrehde:
Maybe this is related to #919?

hankertrix:
Similar problem on iPad: #740

RamsaKd:
> I can't reproduce this, please can you tell me what features are in your note like pdfs, images, shapes, etc?

I've encountered this issue with pdfs with nothing but handwriting on them as well as ones with handwriting, pdfs & images in any possible combination of those 3. But I have also been able to export pdfs with handwriting on them. I have had this issue on linux, grapheneOS & on a samsung galaxy tab s7 fe.

RamsaKd:
I've attached a zip with 2 files, one that successfully exports & one that doesn't. Both are nothing more than handwriting. They were part of 1 multi-page file that I deleted parts of for each version if that is relevant.
[Uploading Export.zip…]()

RamsaKd:
I've attached a zip with 2 files, one that successfully exports & one that doesn't. Both are nothing more than handwriting. They were part of 1 multi-page file that I deleted parts of for each version if that is relevant.
[Uploading Export.zip…]

RamsaKd:
I've attached a zip with 2 files, one that successfully exports & one that doesn't. Both are nothing more than handwriting. They were part of 1 multi-page file that I deleted parts of for each version if that is relevant. [https://drive.proton.me/urls/0JG7TVCF68#QJyFrAUWHiDB](url)

BHuck74:
Hi there!

I've also attached a zip with one document that successfully exports, and the other doesn't. For the one that doesn't export, I've included the .sba file (export to .sba works, only pdf does not).

[...]

PR #1135, opened 2024-02-23
fix: ignore NaN points on pdf export

arpandaze:
While converting points to SVG path, the condition where dx and dy values of points are NaN isn't checked. So path returned by toSvgPath() becomes something like `MNanNaNLNaNNaN`  which is invalid.This is the reason why PDF exports are failing. I have fixed that by filtering out NaN values.

image

Closes https://github.com/saber-notes/saber/issues/1026
Closes https://github.com/saber-notes/saber/issues/1125


adil192:
Thank you!

PR #1163, opened 2024-03-10
i18n: Update Czech

ceskyDJ:


ceskyDJ:
@adil192, just thinking… Maybe you could update all other READMEs, too?

PR #1182, opened 2024-03-18
feat: Customizable toolbar button size

### LINKED ISSUE #1146: Toolbar in two rows after pencil button was added

### Describe the bug

In the new version of Saber 0.20.3 is new toolbar icon for pencil.  On my tablet is now a toolbar in two rows

image

It takes too big area of screen. I suppose that other users have the same issue.

### To reproduce

Install Saber v0.20.3

### Expected behavior

Is it possible to use smaller toolbar buttons and keep toolbar in one row?

### Saber version

v0.20.3 android

### Device

- Device: Galaxy Tab S6
- OS: Android

### Anything else?

_No response_

### LINKED ISSUE #1177: Text Feature ist unusable

### Describe the bug

If I try to enter a text in a note on the S23 using the keyboard, this is not possible because the entire screen is filled with the text attribute window and the keyboard.
See screenshot

threema-20240317-150109551

### To reproduce

Press the TT button on a cell phone within a note

### Expected behavior

The window with the selection of text attributes (bold, italics, etc.) should be collapsible and expandable.

### Saber version

0.21.2 FDroid

### Device

- Device: S23
- OS: Android 14 OneUi 6 

### Anything else?

_No response_

QubaB:
Implemented 3 different sizes of toolbar buttons to allow user to choose smaller buttons if toolbar takes too big area of screen.

Closes #1146  
also partialy closes #1177 (here probably the possibility to hide toolbar when typing text will be better solution - for example long press on text toolbar button)

### DIFF
--- a/lib/components/toolbar/toolbar_button.dart
+++ b/lib/components/toolbar/toolbar_button.dart
@@ -1,81 +1,28 @@
 import 'package:flutter/material.dart';
 import 'package:saber/data/prefs.dart';
 
-/// toolbar button size
-enum ToolbarSize {
-  big,
-  normal,
-  small,
-}
+/// toolbar buttons sizes
+enum ToolbarSize{
+  small(buttonSize: 12, padding: 2, colorOptionDiameter: 18, penModalSize: 22, sizePickerFontSize: 8, sizePickerSize: Size(150,18)),
+  normal(buttonSize: 18, padding: 6, colorOptionDiameter: 28, penModalSize: 32, sizePickerFontSize: 10, sizePickerSize: Size(150,28)),
+  big(buttonSize: 22, padding: 8, colorOptionDiameter: 30, penModalSize: 40, sizePickerFontSize: 14, sizePickerSize: Size(150,30));
 
-extension ToolbarSizeGet on ToolbarSize {
-
-  double getButtonSize(){
-    /// return size of toolbar button
-    switch (this){
-      case ToolbarSize.small:
-        return(12.0);
-      case ToolbarSize.normal:
-        return(18.0);
-      case ToolbarSize.big:
-        return(22.0);
-    }
+  const ToolbarSize({
+    required this.buttonSize,
+    required this.padding,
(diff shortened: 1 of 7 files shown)


lib/components/toolbar/toolbar_button.dart
```diff
@@ -1,4 +1,83 @@
...
+  big,
+  normal,
+  small,
+}
+
+extension ToolbarSizeGet on ToolbarSize {
+
+  double getButtonSize(){
+    /// return size of toolbar button
+    switch (this){
```
adil192, line 15: I've not had time to have a proper look at this but can you refactor to use [Dart's enhanced enums](https://dart.dev/language/enums#declaring-enhanced-enums) instead, like
```dart
enum ToolbarSize {
  big(buttonSize: 22, ...),
  normal(...),
  small(...);

  const ToolbarSize({
    this.buttonSize,
    ...,
  });

  final double buttonSize;
  // ...
}
```

lib/i18n/strings.g.dart
```diff
@@ -521,6 +523,19 @@ class _StringsSettingsLayoutSizesEn {
 	String get tablet => 'Tablet';
 }
 
+// Path: settings.layoutSizes
+class _StringsSettingsEditorToolbarSizesEn {
```
adil192, line 527: I think you forgot to commit `strings.i18n.yaml`

adil192 (changes requested):

lib/components/toolbar/toolbar_button.dart
QubaB: I will try to rewrite code.

QubaB: code is rewriten

lib/i18n/strings.g.dart
QubaB: added strings.i18n.yaml (added manually string for toolbar button size).  Please correct it if i did it wrong.

PR #1187, opened 2024-03-21
fix: Handwriten text is incorrectly recognized as straight line 

### LINKED ISSUE #1186: Writing in small cursive ends up in having the word replaced by a straight line

### Describe the bug

I did a long writing session on my Google Pixel Tablet with GrapheneOS on Saber v0.21.2 F-Droid (21020) with an USI 2.0 Pen (Line thickness "2", Line height 60).

I write mostly in traditional and sloppy German cursive and often without lifting the pen for a whole word.

Everytime I wrote a word that was about 5mm in height and 15mm or more in length with a right ponting starting and end stroke, my word was converted to a straight line.

### To reproduce

The picure shows two examples of writing. Both done on paper and on the tablet. The first one is fine, because it starts with a downward stroke an ends with a stroke to the right. The second one starts and ends with the pen moving in the same direction and gets converted to a staight line.

### Expected behavior

Assistans with straight lines is nice and all, but there is much change in direction in the given example. That schould prevent the algorithm from "helping with straight lines".

### Saber version

v0.21.2 F-Droid (21020)

### Device

- Device: Google Pixel Tablet
- OS: GrapheneOS (Android)

### Anything else?

Thank you very much for developing Saber as an Free Open Source Applikation.

QubaB:
When handwriten text is very small (in y direction) it is incorrectly converted to straight line

Implemented fix checks if pen is not "Shape pen". 

It is tested if x direction of continuous line is not changed during writing. If the direction is changed it indicates that it is not line but rather text.

Line should be at least 3 times longer in x than in y direction

Closes #1186

### DIFF
--- a/lib/components/canvas/_circle_stroke.dart
+++ b/lib/components/canvas/_circle_stroke.dart
@@ -140,7 +140,7 @@ class CircleStroke extends Stroke {
 
   @override
   @Deprecated('We already know the shape is a circle.')
-  bool isStraightLine([int minLength = 0]) => false;
+  bool isStraightLine({int minLength = 0,bool isShapePen=true}) => false;
 
   @override
   CircleStroke copy() => CircleStroke(
--- a/lib/components/canvas/_rectangle_stroke.dart
+++ b/lib/components/canvas/_rectangle_stroke.dart
@@ -161,7 +161,7 @@ class RectangleStroke extends Stroke {
 
   @override
   @Deprecated('We already know the shape is a rectangle.')
-  bool isStraightLine([int minLength = 0]) => false;
+  bool isStraightLine({int minLength = 0,bool isShapePen=true}) => false;
 
   @override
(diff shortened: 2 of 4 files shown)


QubaB:
According to discussion in #1186 the algorithm recognizing hand writen text changed to calculate dot product of start-end point of stroke vector and vector of each segment of stroke - id dot product is <=0 then those vectors are 90° and more.

adil192:
This doesn't work unless you draw the line perfectly. A wobbly line will have lots of perpendicular segments

QubaB:
I assumed that if someone wants to draw a straight line, they draw it rather quickly.

Perhaps it would be possible to determine the total length of the line and the distance between the start and end points as a decision criterion. If the length of the line was multiple times longer than the distance between the beginning and the end, it would be text. Would this approach be acceptable?

PR #1191, opened 2024-03-25
fix: use pdf filepath to cache asset index

### LINKED ISSUE #1189: Saving note causes freeze for 10-60 seconds depending on the imported pdf's size.

### Describe the bug

When saving a note, the entire app becomes unresponsive until saving is completed.

### To reproduce

1. Open a note
2. Import pdf
3. Write/use the note normally
4. Save/autosave
5. Experience a complete freeze of the App for 10-60 seconds (proportional to pdf's filesize)

### Expected behavior

The app shouldn't freeze.

### Saber version

v0.21.2 F-Droid (21020)

### Device

- Device: Samsung Galaxy Tab S6 Lite
- OS: Android 13 (LineageOS)

### Anything else?

v0.21.1 doesn't have this issue.

atahanacar:
Don't compare pdf files byte by byte when adding it to OrderedAssetsCache. This causes performance issues as this operation is done for every page, which gets very expensive with many pages and big pdf sizes.

Cache the index of the asset using the pdf's file path.

Closes #1189


lib/components/canvas/_asset_cache.dart
```diff
@@ -83,6 +83,18 @@ class OrderedAssetCache {
 
   final List<Object> _cache = [];
 
+  /// Maps a file path to already added assets.
+  final Map<String, int> _fileIndices = {};
+
+  void addAsset(File? file, int index) {
```
QubaB, line 89: Better to use addPdfAsset name because this is only for pdf files

lib/components/canvas/_asset_cache.dart
```diff
@@ -83,6 +83,18 @@ class OrderedAssetCache {
 
   final List<Object> _cache = [];
 
+  /// Maps a file path to already added assets.
+  final Map<String, int> _fileIndices = {};
```
QubaB, line 87: better use _filePdfIndices

lib/components/canvas/_asset_cache.dart
```diff
@@ -83,6 +83,18 @@ class OrderedAssetCache {
...
+
+  void addAsset(File? file, int index) {
+    if (file == null) return;
+    _fileIndices[file.path] = index;
```
QubaB, line 91: use _filePdfIndices

lib/components/canvas/_asset_cache.dart
```diff
@@ -83,6 +83,18 @@ class OrderedAssetCache {
...
+  }
+
+ int? getAsset(File file) {
+    return _fileIndices[file.path] as int?;
```
QubaB, line 95: getPdfAsset  and _filePdfIndices

QubaB:
It will be nice to push this fix as soon as possible because the long saving of notes with large pdf files is annoying (#1189).

lib/components/canvas/_asset_cache.dart
QubaB: May be addPdfAssetIndice  will be even better, because we store only indice not asset itself.

lib/components/canvas/_asset_cache.dart
atahanacar: This can be used for non-pdf assets as well. Wouldn't it be better to keep the name?

QubaB: I agree. Please mark my review as closed.

atahanacar:
I think this should be just a temporary fix as pdf file assets are still overwritten on every save. I will work on a better fix when I have the time, as I think there are more important bugs to fix.

QubaB:
I think so. 

I suppose that @adil192 did not expect such great use of large pdfs as I observe now. It seems to me that a lot of users do not use Saber to handwriting of notes but to comment pdfs.

QubaB (changes requested):
Lines 102-115 in lib/components/canvas/_asset_cache.dart
` ``
if (index == -1 && value is List<int>) {
      // Lists need to be compared per item
      final listEq = const ListEquality().equals;
      for (int i = 0; i < _cache.length; i++) {
        final cacheItem = _cache[i];
        if (cacheItem is! List<int>) continue;
        if (!listEq(value, cacheItem)) continue;
        index = i;
        break;
      }
    }
    log.fine('OrderedAssetCache.add: index = $index, value = $value');
```
should be removed. They arecausing slow saving of pdfs by byte to byte comparison.

AddFileIndex and getFileIndex solve this.

atahanacar:
I think they should stay for non-file assets (don't know if they exist, will check when I get back home). We can have another method in OrderedAssetCache that accepts a file as an argument, so it uses the new filepath cache implemented in this PR.

QubaB:
This code was added by me and will be executed only when value is List<int> it means file.  It was introduced because `_cache.indexOf(value);`   sometimes returned -1 index even if value was already in assets (it is probably problem of calculation of hash inside Object class.)

Imagine situation of two images (pdf files) in a note. When index == -1 (it means that asset is not in cache yet) image bytes will be compared to all already stored assets byte by byte which is slow.  But your fix does not allow to call  assets.add when it already knows ID of asset.

So in case of more images each new image will be at least once compared to all previously stored images.

Please test your code with more than one pdf files imported.

I hope my explanation is clear.

atahanacar:
Yes, it is very clear. Thanks. I will do more testing and see how I can safely remove your code without causing the same issue you fixed. However, in the meantime I think this PR should be merged as soon as possible.

Also as a funny side note, I've been wanting to begin contributing code to this project for a long time but was procrastinating. The long saving bug your fix introduced actually made me finally start it, so I want to thank you for that 🤣

QubaB:
> The long saving bug your fix introduced actually made me finally start it, so I want to thank you for that 🤣

I am glad you started contributing.  I am a FORTRAN programmer of scientific calculations and my contributions are clumsy. Someone else will do it better.

atahanacar:
@adil192 please merge this as soon as possible, this is a critical issue that affects all platforms. I can't use the official or F-Droid builds because of this.

Luciogi:
@adil192 any thoughts on this ?

QubaB:
Please @adil192 consider to merge this branch. It improves time of saving note when it contains more large pdf files (avoids comparing each pdf file byte by byte).

atahanacar:
No longer necessary since 5584b30

PR #1192, opened 2024-03-26
feat: background line thickness settings added

arpandaze:
Adds settings to control the thickness of background lines.


adil192 (changes requested):
This is cool, can you just refactor `lineWeight` to `lineThickness`. Atm it looks too similar to `lineHeight`!

arpandaze:
> This is cool, can you just refactor `lineWeight` to `lineThickness`. Atm it looks too similar to `lineHeight`!

Sure. I have just refactored it into lineThickness but pipeline seems to be failing due to late initialization in 1 test. I have also added default value for the preference but not sure why its throwing error. Can you please let me know how to fix this?

adil192:
> Sure. I have just refactored it into lineThickness but pipeline seems to be failing due to late initialization in 1 test. I have also added default value for the preference but not sure why its throwing error. Can you please let me know how to fix this?

Oof yeah that was a tricky issue.
The reason was that large files (like v6_stress_test_179.sbn) are parsed on a separate isolate which doesn't have access to Prefs.
I fixed it by falling back to the default line thickness instead of the last line thickness if we're not on the main isolate

PR #1197, opened 2024-03-30
feat: choose custom data directory on Android

### LINKED ISSUE #24: Allow Android users to sync to a local folder
Would you consider adding Syncthing support? Nextcloud is too heavy-weight for my home use case, where Syncthing suffices with its P2P support that doesn't require internet connection. It should be fairly easy to implement. [Joplin](https://github.com/laurent22/joplin-android) does this by synchronizing notes to/from a directory in the system.

adil192:
Resolves #24

### Todo
- [X] Find out if [`Directory.rename(...)`](https://api.flutter.dev/flutter/dart-io/Directory/rename.html) does what I want (it didn't)
- [X] Test on a real device

### DIFF
--- a/lib/components/settings/settings_directory_selector.dart
+++ b/lib/components/settings/settings_directory_selector.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:file_picker/file_picker.dart';
 import 'package:flutter/cupertino.dart';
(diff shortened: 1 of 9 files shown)


imadnyc:
Could test if needed -- this is one of my most anticipated features!

PR #1213, opened 2024-04-07
fix: validate name after dispose

ZebraVogel94349:
I've noticed that validation of the new name when renaming a note in the editor doesn't work, so you can rename a note to an empty name or a name containing slashes from the editor. The problem was that `_filenameFormKey.currentState?.validate() ?? true` is always a string or true, but it should be false if the function returns null. I've also fixed a bug which made it impossible to save a note after renaming it in the editor but before the new name has been applied.

### DIFF
--- a/lib/pages/editor/editor.dart
+++ b/lib/pages/editor/editor.dart
@@ -921,7 +921,8 @@ class EditorState extends State<Editor> {
     final newName = filenameTextEditingController.text;
     if (newName == coreInfo.fileName) return;
 
-    if (_validateFilenameTextField(newName) == null) {
+    if (_filenameFormKey.currentState?.validate() ??
+        _validateFilenameTextField(newName) == null) {
       coreInfo.filePath = await FileManager.moveFile(
           coreInfo.filePath + Editor.extension, newName + Editor.extension);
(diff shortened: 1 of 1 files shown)


adil192:
> _filenameFormKey.currentState?.validate() ?? true is always a string or true

The [FormState.validate] function returns true if there are no errors, otherwise false. Dart wouldn't accept an "if (string)" anyway.

> I've also fixed a bug which made it impossible to save a note after renaming it in the editor but before the new name has been applied.

I'll need to look at this further but I don't have time rn

ZebraVogel94349:
> The [FormState.validate] function returns true if there are no errors, otherwise false. Dart wouldn't accept an "if (string)" anyway.

You're right, that doesn't solve the issue. I misunderstood how the validate() function works.

ZebraVogel94349:
I've fixed it now. When you exit the editor after renaming the file, but before it has been saved, `_filenameFormKey.currentState` is null, so `_filenameFormKey.currentState?.validate() ?? true` is true and the file will be renamed, even though the validation didn't happen. By calling `_validateFilenameTextField(newName)` directly, the name is always validated.

PR #1214, opened 2024-04-08
i18n: Update Czech

ceskyDJ:


ceskyDJ:
@adil192, what about making an announcement for other translation contributors?

adil192:
Thanks!

Issue #1228, opened 2024-04-22
Notes become unaccessible when moving fails


ZebraVogel94349:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

I just updated Saber to 0.21.4 and changed directory where notes are stored on Android. In one folder I had a glitched note with the name '/', which got created due to the bug described in #1213. Because of that, I know have a bunch of folders containing other folders as my Saber directory and can't access any of my notes, which are still in /data/data/com.adilhanney.saber/ and didn't get moved. Please make it possible to change the saber directory back to the default one, even though it still has notes in it.

### To reproduce

1. Create a note with the name '/' by renaming it in the editor
2. Go to settings and change the saber directory
3. The app tries to recursively copy all notes, resulting in the app to crash
4. The new directory doesn't contain any notes. They are as far as I know still in the default one.

### Expected behavior

Don't move anything named '/'.

### Saber version

v.0.21.4

### Device

- Device: Samsung Galaxy Tab S7 FE
- OS: Android 14

ZebraVogel94349:
I just made a quick fix to this by making it possible to change the directory to one that is not empty and then just not moving anything. The fix is in #1229

PR #1229, opened 2024-04-22
fix: allow selecting new directory that is not empty

ZebraVogel94349:
I know this solution is not ideal, but it makes it possible to recover notes that got lost due to #1228.


ZebraVogel94349:
I closed this pull request because notes can be recovered by downgrading the app via adb, so this fix is not needed anymore.

PR #1238, opened 2024-05-02
i18n: Update READMEs

ceskyDJ:


ceskyDJ:
@adil192 and maybe it'll be good to propagate the change to other language mutations of README, too.

adil192:
Thanks!

PR #1250, opened 2024-05-07

jhihyulin:


adil192:
Thanks for the speedy translation! I've since changed this functionality though so I'll have to close this PR

PR #1263, opened 2024-05-15
snap: add gstreamer libraries

soumyaDghosh:


adil192:
We may also need `libgstreamer-plugins-base1.0-dev` for ogg support, but I'm not sure if that's necessary here?

soumyaDghosh:
> We may also need `libgstreamer-plugins-base1.0-dev` for ogg support, but I'm not sure if that's necessary here?

The libraries are in the runtime sdk, but the pkgconfig and development files got removed in the sdk, thus the need of these files. So, we need files mostly to satisfy the build. I'll try to look into it once again.

adil192:
Please mark this as "ready for review" when you're ready

adil192:
I'll just merge this for now to try and get builds working

PR #1268, opened 2024-05-19
i18n: Update Czech

ceskyDJ:
Hi, @adil192! I'm here with some fixes as I promised in the last pull request!

There is a few things I'd live to fix but I simply can't do it:

1. Some languages (like Czech) uses decimal comma instead of decimal dot (1.21 is correctly 1,21 in Czech). Maybe it could be done just by setting locale in the Flutter or something like that, I don't know.
2. We use a little more strict rules for percentage character. Simply said: we have two meaning for the % char — the meaning without space (21%) is different than meaning with space in between (21 %). So, if you could replace `$percent` with just the number, it will be nice. I need to have usage description typeset as "21 %" (without quotes) for Czech.
3. Czech adds a (unbreakable, but it's not necessary in the context of the mobile app) space between number and its unit (1 s, for example). I found that in app settings, numbers and units are connected together. It has a little different meaning in Czech. Could you divide the number and the unit in the translation string, please?

### DIFF
--- a/lib/i18n/community/strings_cs.i18n.yaml
+++ b/lib/i18n/community/strings_cs.i18n.yaml
@@ -146,7 +146,7 @@ settings:
     mustBeEmpty: Zvolený adresář musí být prázdný
     mustBeDoneSyncing: "Než změníte datový adresář, ujistěte se, že byla dokončena synchronizace"
 login:
-  title: Přihlásit se
+  title: Přihlášení
   form:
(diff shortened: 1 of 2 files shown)


ceskyDJ:
@adil192, it looks like we're stuck here a bit. Could you look at it, please? If it's waiting due to stuff I proposed in my comment, I think we can move it to separate issue and merge this, can't we?

PR #1274, opened 2024-05-22
perf: reduce pencil sound effect latency

adil192:


adil192:
The latency is actually worse with this package. About half a second

Issue #1304, opened 2024-07-02
Unable to sync with saber cloud


JustDj-Git:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

Unable to sync with saber cloud. I tried to:
-create a new acc
-change devices (tried on Samsung tab s9, tab s6, nothing phone 1 and 2, appimage and flatpak versions on Linux
-reinstall saber everywhere

sometimes files are synced with cloud (I can see them in cloud) but if I delete notes from my pad/phone - I cant resync them

Error on Linux (flatpak):
`flutter: WARNING: SaberSyncInterface: Failed to get sync file from remote file: Exception: Decryption failed for Saber/
`
`flutter: WARNING: SaberSyncInterface: Failed to get sync file from remote file: Invalid argument(s): Invalid or corrupted pad block`

### To reproduce

1. Sign-in on tab + phone (or more devices)
2. Create any note and sync
3. Try to sync notes on other devices - nothing happens 

### Expected behavior

Sync :)

### Saber version

v0.24.1

### Device

- Devices: Linux PC, Samsung s9 tab, Samsung tab s6, nothing phone 1 and 2
- OS: Debian, Arch, OpenSUSE, Fedora, Android

TekkertheChaot:
same here! have the same bahaviour using the flatpack 0.24.1 on linux and 0.24.1 on android. 
I also saw the same journal entry described above.

I've attached a video reproducing this behavior
https://github.com/saber-notes/saber/assets/17234581/10b83ebd-3871-45e2-804c-d9931bc8bb7a

gewinh:
Same here Saber v0.24.1 syncs to the nextcloud server with what looks like the right file size, but there is no syncing between my Android devices. I have a phone and a tablet each has the same version of saber and each is setup to sync to the same Nextcloud account with the same encryption password. They both sync there files to the saber folder but files created on the phone don't show up on the tablet and viz versa. I have also installed Saber on ubuntu as a snap and have the same issue, no syncing between devices. I am running Nextcloud 29.02 behind an nginx reverse proxy. Other apps that are using nextcloud sync are working fine. eg. Moonreader, files, notes deck etc.

[...]

PR #1314, opened 2024-07-17
fix: remote files are not synced

### LINKED ISSUE #1306: Apps won't sync with each other (Ipad and linux)

### Describe the bug

I have both the linux version (flatpak) and the ipad os version
i'm using my own nextcloud server, and i can see files in the saber dir:
image
When i create a new file it happers in the nextcloud folder, but i can't see it in other clients eg: files created in the ipad app won't appear in the linux app and viceversa:
Linux app:
image
ipad app:
Screenshot 2024-07-08 (00 12 17) jpeg

i've double checked and i'm using the same encryption password in both instances.

### To reproduce

1. create note on device 1
2. create note on device 2
3. See them appear on the nextcloud folder (they can successfully sync)
4. note that was created on device 1 won't appear on device 2 and viceversa

### Expected behavior

Beeing able to see and edit notes from both devices

### Saber version

Linux:
0.24.1
Ios:
0.24.1 (AppStore)

### Device
- Device 1 : Ipad 10
- OS: IpadOS 17.5.1

- Device 2 : my main desktop
- OS: Linux Mint 21.3

QubaB:
fix problems with syncing notes from nextcloud server

closes #1306

### DIFF
--- a/lib/data/nextcloud/saber_syncer.dart
+++ b/lib/data/nextcloud/saber_syncer.dart
@@ -78,10 +78,10 @@ class SaberSyncInterface
       final bestFile = await getBestFile(syncFile);
       switch (bestFile) {
         case BestFile.local:
-        case BestFile.either:
           // Local file is newer, do nothing
           break;
         case BestFile.remote:
+        case BestFile.either: //local file does not exists
           changedFiles.add(syncFile);
       }
     }
--- a/lib/data/file_manager/file_manager.dart
+++ b/lib/data/file_manager/file_manager.dart
@@ -166,7 +166,13 @@ class FileManager {
     } else {
       assert(filePath.startsWith('/'),
           'Expected filePath to start with a slash, got $filePath');
-      return File(documentsDirectory + filePath);
+      if (filePath.startsWith(documentsDirectory)) {
+        // already contains document directory path
+        return File(filePath);
+      }
+      else {
+        return File(documentsDirectory + filePath);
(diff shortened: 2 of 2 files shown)


QubaB:
closes  #1304 also

adil192:
Thanks for finding this! I've not been able to for ages.
However, this is just hiding the problem which is that the argument to `FileManager.getFile` should be the local relative path not an absolute path. I'm going to address the root issue in a separate commit and give you credit.

lib/data/nextcloud/saber_syncer.dart
```diff
@@ -78,10 +78,10 @@ class SaberSyncInterface
       final bestFile = await getBestFile(syncFile);
       switch (bestFile) {
         case BestFile.local:
-        case BestFile.either:
           // Local file is newer, do nothing
           break;
         case BestFile.remote:
+        case BestFile.either: //local file does not exists
```
adil192, line 84: Why did you change this part? `BestFile.either` should be when the local and remote files are equivalent, so no sync needs to occur for this file

QubaB: BestFile.either is also returned when local file does not exist. I did not searched the code why.

adil192: Thanks, I've improved this aspect in 01b28d9fb2f62672849de58841c573ebad908443

PR #1322, opened 2024-07-25
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Hi @adil192, I noticed the new string for `customDataDir` is a little confusing. I translated it a bit different as "Vlastní umístění složky aplikace Saber" (Custom location of Saber folder) because there is a collision between "custom folder" and "Saber folder".

PR #1355, opened 2024-09-19
feat: Add better support for stylus hovering

### LINKED ISSUE #859: Pen buttons on Android
### Describe the bug

A button on the pen, it's usually for erasing, or whatever you set it.

### To reproduce

1. Open the app
2. Start a new note or whiteboard
3. Draw
4. Press the button on the pen

### Expected behavior

When you press the button, it should select the eraser or hilghliter tool; relase the button, and 6oure back to pen input.

### Saber version

v0.14.11 Google Play (14110)

### Device

- Device: Samsung Galaxy Fold 4
- OS: Android 13

### Anything else?

Great app, but please record yourself drawing with a pencil on paper and add it as an additional option to hear the brushing. Otherwise, responsive, still great!

3akev:
This PR adds handlers for hovering stylus events, which are used to switch to the eraser tool when the stylus button is pressed while hovering.

This replaces the old functionality where pressing the stylus button mid-stroke removes the stroke, making it more in-line with other note applications and (in my opinion) more user friendly.

Closes #859

### DIFF
--- a/lib/components/canvas/canvas_gesture_detector.dart
+++ b/lib/components/canvas/canvas_gesture_detector.dart
@@ -392,25 +392,35 @@ class CanvasGestureDetectorState extends State<CanvasGestureDetector> {
 
   void _listenerPointerEvent(PointerEvent event) {
     double? pressure;
-    bool stylusButtonPressed = false;
 
     if (event.kind == PointerDeviceKind.stylus) {
       pressure = event.pressure;
-      stylusButtonPressed = event.buttons == kPrimaryStylusButton;
     } else if (event.kind == PointerDeviceKind.invertedStylus) {
       pressure = event.pressure;
-      stylusButtonPressed = true; // treat eraser as stylus button
     } else if (Platform.isLinux && event.pressureMin != event.pressureMax) {
       // if min == max, then the device isn't pressure sensitive
       pressure = event.pressure;
     }
 
     widget.onPressureChanged(pressure);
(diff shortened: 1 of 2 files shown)


adil192:
This looks good from the code. What device/s have you tested with? iirc pressing the S-pen button while hovering made the stylus not work temporarily

3akev:
> This looks good from the code. What device/s have you tested with? iirc pressing the S-pen button while hovering made the stylus not work temporarily

I tested this on a Galaxy S6 Lite with the S-pen. Unfortunately I don't have access to another device with a stylus, so I can't test this further. It should work on other devices, though.

adil192:
Thanks!

QubaB:
On my Galaxy Tab S6 is tool changed to Eraser when pressed button, but entities are not erased. So it does not work.

Does erasing work for you? I am using official relase 0.25.3.

3akev:
> On my Galaxy Tab S6 is tool changed to Eraser when pressed button, but entities are not erased. So it does not work.
> 
> Does erasing work for you? I am using official relase 0.25.3.

I just tested the latest release, 0.25.3 on Galaxy S6 Lite. It appears to be working. Seems this is a rather sneaky bug. Maybe open an issue for it?

QubaB:
I have found, that Preference "Automatically switch off Eraser"  (do not know precise English name) does not work on my Saber 0.25.3. Even if this preference is switched On or Off, previous tool is used immediately after use of Eraser. I do not know if it can be connected.

PR #1358, opened 2024-09-26
i18n: Update Czech

ceskyDJ:
Hi @adil192!

It was very nice to see Saber has enough sponsors to pay for its own bills! I hope there will be more of them, so you will be motivated to maintain this project.

I've updated Czech translation of README. There were some minor mistakes I wanted to fix.


adil192:
Thanks!

PR #1377, opened 2024-10-25
feat: Added select/deselect all notes button

Edgiest05:
Introduced button in recent page footer that allows for selection of all notes

When less than all files are selected
immagine

When all files have already been selected, behaviour changes to unselect all
immagine

While implementing this I had to address that newer version of the `saver_gallery` dependency would not compile.
Furthermore `preview_card` makes use of a different variable `expanded` when it actually should never differ from the `widget.selected` one (or else some selection changes do not reflect in the UI). I've kept it for consistency's sake.

### DIFF
--- a/lib/components/home/preview_card.dart
+++ b/lib/components/home/preview_card.dart
@@ -78,6 +78,7 @@ class _PreviewCardState extends State<PreviewCard> {
 
   @override
   Widget build(BuildContext context) {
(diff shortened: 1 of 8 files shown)


Edgiest05:
Same functionality is now available also in each directory of the browse page

Edgiest05:
Resolved merge conflicts and updated `saver_gallery` to new version.

PR #1408, opened 2024-12-01
feat: Sort notes and folders in recent and browse pages

Edgiest05:
Introduced a button `SortButton` located at the top-right of the UI in recent and browse pages (near the syncing one):
button

When pressed it shows a dropdown for a predetermined selection, provided in the static class `SortNotes`.
Such selection is preserved as a preference and is kept on app restart.
dropdown

Sdk version change is not related to my own implementation, but instead is a workaround to compilation errors that made it impossible to test on android emulator.

### DIFF
--- a/android/app/build.gradle
+++ b/android/app/build.gradle
@@ -52,7 +52,7 @@ android {
         applicationId "com.adilhanney.saber"
         // You can update the following values to match your application needs.
         // For more information, see: https://docs.flutter.dev/deployment/android#reviewing-the-build-configuration.
-        minSdkVersion 23
+        minSdkVersion 24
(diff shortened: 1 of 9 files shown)


QubaB:
Thank you for Your code, I merged it to my personal Saber.

QubaB:
I like to sort folders by Name and files by date. Can this be implemented?

Edgiest05:
@QubaB it is possible to implement and it is quite easy to do so given that I already handle files and directories separately when sorting, unfortunately I personally don't need this level of fine tuning and I probably won't be working on that

QubaB:
OK, I will customize it to my needs. Thank you for the very nice improvement.

PR #1428, opened 2025-02-16
fix. Stylus button press changes tool to Eraser but do not erase.

QubaB:
Some styluses do not send move events when button is pressed. So Eraser do not work.

This change allows switch to eraser by pressing and releasing stylus button (no more holding button), then erasing items work.

Creates new preference, which can change default behavior of pressing button when pen is above screen.

Solves issue saber-notes#1410


adil192 (changes requested):
@QubaB Thanks for the fix.

Please add some tests for this behaviour with and without the eraserOnStylusButtonPressAndRelease setting.

Also, the setting is a bit confusing at the moment: I think the most understandable UI would be to have two selection options (in an  enum) for "Press stylus button to toggle eraser" and "Hold stylus button down for eraser".

PR #1430, opened 2025-02-19
fix: Synchronizing file from nextcloud does not set correct time stamp

QubaB:
When file is synchronized from remote (Nextcloud) to local, set file time stamp according to time stamp of remote file. This avoids problem than newly created note is immediately synchronized to nextcloud, because it has newer time stamp.

Another change:
In the file name are replaced '\' to '/'  before syncing to nextcloud. Windows use '\' and such file is treated as different file than '/'.  It caused problems with synchronization between windows and Android.


adil192:
Working for me! thank you so much

Issue #1336, opened 2024-08-11
Windows-Client does not store Nectcloud-Connection (URL) or key


thomas-unk:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

I experience a problem with Windows Saber client (Windows 11 Pro). I can set it up and connect to nextcloud. Synchronization works. But with each restart of Saber, I am requested my nextcloud server settings (URL) and I have to re-confirm cloud access by logging into the nextcloud instance. It does not store URL or key.

### To reproduce

Start Saber
Connect with your own Nextcloud
End Saber
Start Saber
Search for the Connection

### Expected behavior

Saber stores the Nextcloud-Connection

### Saber version

v0.24.5 (Windows)

### Device

- Device: HP EliteBook 
- OS: Windows 11 Pro

Obi-Wahn:
probably the same problem as mine. it is a bug with flutter_secure_storage: https://github.com/saber-notes/saber/issues/1252

pisceswb:
+1

osiktech:
I still experience this in my Windows 11 Pro ARM64 VM. The file `flutter_secure_storage.dat` is created but I always have to login. on my MacOS and Androids I do not have this issue. I am using Saber v0.25.9 (25090)

QubaB:
My fix is not accepted yet. Problem is only on Windows.

Obi-Wahn:
> My fix is not accepted yet. Problem is only on Windows.

Thank you for your efforts to solve the problem! 👍🏼

PR #1431, opened 2025-02-20
fix: Storing Nextcloud credentials on Windows does not work

QubaB:
This fix changes how the Nextcloud credentials are stored using flutter_secure_storage.  

The main reason saving credentials did not work on Windows is that there was no wait after saving the change to secure storage.

I tested this fix on Windows and Android. It was not tested on iOS.

Solves #1336


adil192:
> there was no wait after saving the change to secure storage

Calling `pref.waitUntilSaved()` doesn't do anything besides check a flag, it should still save in the background. The Prefs (now called Stows) system was designed so we don't need to await after a get/set.

Was `pref.waitUntilSaved()` the fix itself? Please can you summarize the changes you made and why they were needed? There are a lot of lines of code, some unnecessary, that were changed in this PR so it's difficult to review without more context.

QubaB:
I created pull request to Stow class which solves this problem [https://github.com/adil192/stow/pull/1](https://github.com/adil192/stow/pull/1).  

There is no need to import my pull request in Saber. Fix of stow is enough.

PR #1434, opened 2025-02-21
fix: switch appimage to new ubuntu version

Drsheppard01:
Using a fresh build should help with glib issues like:

```/tmp/.mount_Saber-7mvgGu/saber: symbol lookup error: /lib64/libgsttranscoder-1.0.so.0: undefined symbol: g_once_init_leave_pointer```

Ubuntu 24.04 uses version 2.80, the previous (22.04) version was 2.72


adil192:
Thanks!

PR #1436, opened 2025-02-23
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Hi, @adil192! It's a pretty long time since there was a last change in i18n!

I have some notes on this translation:

- At first, I translated the description a bit differently, so it's a little more informative (at least from my point of view). Instead of your version, I used something like “Background template's lines thickness”. There is a little problem with the meaning of “line“ in Czech, so I used two forms to describe it more precisely (to add more context for the user).
- The second note is about a preview. Line height (option just above this one) has a real-time preview, but there is nothing like that for line thickness. Could you please add it?

adil192:
> The second note is about a preview. Line height (option just above this one) has a real-time preview, but there is nothing like that for line thickness. Could you please add it?

Tested now and the line thickness is updated in the real-time previews too. Maybe it was fixed since this PR was made

adil192 (approved):
Thanks!

ceskyDJ:
> > The second note is about a preview. Line height (option just above this one) has a real-time preview, but there is nothing like that for line thickness. Could you please add it?
> 
> Tested now and the line thickness is updated in the real-time previews too. Maybe it was fixed since this PR was made

It still doesn't work for me. I have Saber v0.25.4.

PR #1445, opened 2025-03-06
Change github ref of `receive_sharing_intent` to a fork which still exists

frothedoatmilk:
Looks like https://github.com/ad-angelo/receive_sharing_intent has vanished! Until @KasemJaffer does another release, I couldn't build this package on linux without switching this ref. It might be worth hosting our own fork so this doesn't happen again?


adil192:
I just switched back to the upstream repo to fix this

PR #1448, opened 2025-03-13
Fix: Saving files in custom data dir not working on Android

### LINKED ISSUE #1223: Saving files in custom data dir not working

### Describe the bug

On the newest Saber version, custom data dirs selected via the UI (in this case a Syncthing directory) do not allow notes to be saved, and thereby the editor cannot be exited without closing the app manually.

### To reproduce

1. Be on the newest (v0.21.4) version of Saber on Android
2. Set a custom data directory via the Settings UI
3. Create a note and attempt to save it by either backing out or by clicking the save icon at the top left

### Expected behavior

The file to save and the editor to close as normal

### Saber version

v0.21.4

### Device

- Device: Samsung SM-X510
- OS: Android 14 (OneUI 6.0)

### Anything else?

_No response_

ComputerElite:
This pull request fixes not being able to save or see certain notes when saving in a custom data directory by requesting the MANAGE_EXTERNAL_STORAGE permission when you select a custom data directory.

This way notes can be synced via Syncthing and more.

Closes #1223 
Replaces #1442


lib/components/settings/settings_directory_selector.dart
```diff
@@ -114,15 +115,52 @@ class _DirectorySelectorState extends State<DirectorySelector> {
 
     final dir = Directory(directory);
     _directory = directory;
-    _isEmpty = dir.existsSync() ? dir.listSync().isEmpty : true;
+    _isEmpty = (dir.existsSync() ? dir.listSync().isEmpty : true) || true;
```
adil192, line 118: Why was this changed?

DelightReza:
When is it going to merge

lib/components/settings/settings_directory_selector.dart
ComputerElite: Good question, I suppose this could be reversed

Forceizer:
When will this be merged?

adil192:
Unfortunately MANAGE_EXTERNAL_STORAGE generally [isn't allowed on the Play Store](https://support.google.com/googleplay/android-developer/answer/10467955)

Iey4iej3:
> Unfortunately MANAGE_EXTERNAL_STORAGE generally [isn't allowed on the Play Store](https://support.google.com/googleplay/android-developer/answer/10467955)

If you decide to avoid this permission, I guess it better to disable "Custom Saber folder" on Android for now — it is broken without that permission.

aarron-lee:
Apologies for the drive-by comment, and perhaps this is a dumb question, but is there no way to restrict the `MANAGE_EXTERNAL_STORAGE` permission to the F-droid build somehow?

I'm currently actively using the fork that was posted with the working Custom directory, it'd be a shame if it was not possible to have access to custom directories.

2EricBrunet:
> Apologies for the drive-by comment, and perhaps this is a dumb question, but is there no way to restrict the `MANAGE_EXTERNAL_STORAGE` permission to the F-droid build somehow?
> 
> I'm currently actively using the fork that was posted with the working Custom directory, it'd be a shame if it was not possible to have access to custom directories.

I am in the same situation ( [discussions#1447](https://github.com/saber-notes/saber/discussions/1447) ).
(Galaxy Tab 6S Lite with Saber **0.25.9** from F-droid)

ComputerElite:
As I'm using saber myself for school (hence the PR). I could keep my fork synchronised with upstream so you can use that for the time being.

2EricBrunet:
> As I'm using saber myself for school (hence the PR). I could keep my fork synchronised with upstream so you can use that for the time being.

So I resume with my understanding of the situation:
* future version of Saber doesn't read custom directory (for me! `/storage/emulated/0/Documents/`) for permissions not allowed by Google store.
* only solution: default directory of the app (=for me : `/data/user/0/com.adilhanney.saber/app_flutter/Saber`)
* advice to backup: use Syncthing App with default directory
It's correct?

aarron-lee:
Hopefully it will be possible to add this to the F-droid or Github builds. For now, I guess I'll compile an unsigned apk for my own personal usage.

If anybody wants a clean diff that cleanly applies onto the latest upstream main branch, see [custom-dir.txt](https://github.com/user-attachments/files/21316903/custom-dir.txt). `git apply custom-dir.txt`

Iey4iej3:
> * advice to backup: use Syncthing App with default directory

It would require root permission for Syncthing, which is unsafe. I am using several Apps which are also on Play Store, and succeed to access external storage, such as [Trime](https://github.com/osfans/trime). I do not know how they manage to deal with this. It might be that the way Saber accesses external storage requires `MANAGE_EXTERNAL_STORAGE`, but there is a way to avoid it.

aarron-lee:
after checking documentation more, looks like the currently accepted way to implement this is via: [ACTION_OPEN_DOCUMENT_TREE](https://developer.android.com/training/data-storage/shared/documents-files#grant-access-directory)

from what I can tell, `MANAGE_EXTERNAL_STORAGE` is considered legacy

Iey4iej3:
> after checking documentation more, looks like the currently accepted way to implement this is via: [ACTION_OPEN_DOCUMENT_TREE](https://developer.android.com/training/data-storage/shared/documents-files#grant-access-directory)
> 
> from what I can tell, `MANAGE_EXTERNAL_STORAGE` is considered legacy

I guess that you could open an issue proposing this re-implementation.

aarron-lee:
after a bit more investigation, looks like a combination of `ACTION_OPEN_DOCUMENT_TREE` and [contentResolver.takePersistableUriPermission](https://stackoverflow.com/questions/71099575/should-i-release-persistableuripermission-when-a-new-storage-location-is-chosen) would work.

Iey4iej3:
Let's continue at https://github.com/saber-notes/saber/issues/1223

PR #1476, opened 2025-06-10
i18n: Update Czech

ceskyDJ:


ceskyDJ:
My goodness, so many typos in the commit message. It should be ok now. Localization should be correct, I at least triple-checked it 😄.

adil192:
Thanks!

Issue #1455, opened 2025-03-25
Create space between lines


serackhlx:
### Before submitting an issue, please check the following

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Motivation

I want to be able to insert space between lines I've already written using the pencil. There is often the time when you get a new idea that you want to write down, but it is a sub topic of another note. As of now I have to just append the new idea somewhere at the bottom. And then am scrolling back and forth. I know about inserting a new page, but this is not quite the same.

### Solution

It would be cool to a button or some context action to just insert some blank new space and move everything after it a bit down, just like in Apple Notes or similar tools. Other than that I really like this app.

### Saber version

0.25.4 App Store (25040)

### Device

- Device: iPad Air sth.
- OS: iPad OS

QubaB:
You can use selection tool, select items to be moved, and move it down and create empty space. 

Currently there is bug that the selection tool moves selected items but they are not moved on screen. There is pull request #1427 solving this bug but is not accepted yet.

serackhlx:
> You can use selection tool, select items to be moved, and move it down and create empty space.

Thanks for the hint, but in my case I would need to select everything below a certain line. This would mean selecting pages of items and text to move them down. This is not very convenient I believe.

serackhlx:
If some line will be splitted in two pages, then so be it. I don't mind do a bit mind bending to read some separated text. Alternatively one could "just" cut the line in 2 parts and place one part on the current page and the other part on the next page. But I also don't mind the idea of having not equal sized pages. I personally don't even need the pages concept at all. I could live with only one large page.

QubaB:
I prepred pull request with possibility to insert free space. However main Saber branch is now in "development phase" instead Prefs is implemented Stow. It does not work for me. But If you test pull request with saber 29.5 and only up to commit #1126938f607fd3b8c4b91e9b6169fccbc7b68aee  (without main branch merge) it should work.

didierm:
This missing feature is nicely handled in https://github.com/styluslabs/Write , which provides for both horizontal and vertical space insertion / page enlargement.

Thanks for your (meanwhile stale ?) patch, @QubaB ; it's a pity these efforts to add missing features seem to not receive any (positive or negative) feedback by the main developer (who has developed a quite high quality application).

PR #1487, opened 2025-07-02
Insert/remove vertical space on page. Implemented new pen "Insert Pen".

QubaB:
Implemented new pen "Insert Pen". When stroke is created using this pen, all items on page below first point of stroke are moved up/down by the vertical length of stroke.

obrazek

Can be used to create free vertical space on page (or remove it).

When moved items do not fit to page, its height is enlarged to hold all items

Solves #1455


adil192 (changes requested):
Please add some tests and remove the draft label if it's ready for review

PR #1492, opened 2025-07-16
mp3 version of Pencil sound for Windows OS

QubaB:
Added mp3 version of pencil sound. On Windows is ogg nonstandard audio format.


adil192:
> On Windows is ogg nonstandard audio format

Does the ogg not work or is this just a preference?
Mp3 is generally a bad format for looping audios because it has silent gaps at each end.

QubaB:
On my Windows (I have not installed any additional codecs) is ogg format unsupported. Wav can be used instead mp3 if it is better.

Issue #1504, opened 2025-08-11
sba file import now work on Windows


QubaB:
### Before submitting an issue, please check the following

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

Import of sba note does not work on Windows

### To reproduce

1. Try to import sba file on Windows

### Expected behavior

It should work

### Saber version

0.26.2

### Device

- Device:
- OS: Windows 11

PR #1505, opened 2025-08-11
fix: sba note import on Windows

QubaB:
fix: on windows is not possible to import sba file because. Windows use '\\' in file path but Saber expect file path to contain '/' so the new note file name is not created correctly. 

fixes: #1504


adil192:
Thanks!

PR #1506, opened 2025-08-14
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Hmm, it looks like GitHub has just shot itself in its leg (if it has some, there is just one hand in the logo 😄). For some reason, web rebase isn't signed by GitHub, so it seems to be Unverified 🤔. Is it ok with you, or should I sign it myself?

adil192:
It's okay, it shows as Verified when I squash-merge it anyway. Thank you!

Issue #1266, opened 2024-05-19
Can't import some PDFs. 


bryanpaget:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

When I create a PDF with some apps (Markor for example), I can't import them to Saber. When I take the same PDF and import it into Samsung Notes it works, and I am able to export from Samsung Notes and then the PDF will open in Saber.

### To reproduce

1. Export from Markor. 
2. Import into Saber. 
3. Fail. 

### Expected behavior

I expected to be able to import any PDF. 

### Saber version

v0.23.0 Fdroid 23000

### Device

- Device: Samsung tab s9 
- OS: android 14

### Anything else?

Love the app.

bryanpaget:
The problem persists for version 0.23.1.

QubaB:
Can you share small pdf which you cannot import as an example?

bryanpaget:
Created by Android print (does not import to Saber Notes):
[Saber.PDF](https://github.com/saber-notes/saber/files/15435835/Saber.PDF)

Created by Samsung Notes (Saber Notes can import this file, it's the same file as above, simply imported to Samsung Notes and then exported):
[SaberPDF_240524_072120.pdf](https://github.com/saber-notes/saber/files/15435834/SaberPDF_240524_072120.pdf)

QubaB:
I could import Saber.PDF to 0.23.2 (Android) and 0.23.0 (Android emulator). Is it not a problem on your side?

docentYT:
I cannot import any PDF - saber displays just white pages.
Tested on 0.26.3 (and on backported 0.26.10).

QubaB:
pdf cannot be imported as an image (even if it is shown in file picker). It can be imported only as background image or using "Import note" when creating new note.

docentYT:
This is what I do. Importing via "Import note" or as a background doesn't work.

QubaB:
On Android should be pdfs imported (at least pdfs with less than lets say 40 pages). Try to import few pages pdf.

If you are skilled try to use my pull request #1559 where I am suggesting complete rework of Saber asset cache. It should solve problems with pdfs (I am using it in my personal version of Saber already and I am able to display 800 pages pdf).

PR #1510, opened 2025-08-25
fix: allow importing files with uppercase extensions

### LINKED ISSUE #1467: Import filetype check is case sensitive

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

Hey,

I realized today while importing a PDF file in the Android App that Saber seems to check for compatible extensions via case sensitive string matching. If I have a `example.pdf` it will work, but  `example.PDF` gives an error as unsupported extension. After renaming the file to lowercase `pdf` it works.

I might be able to fix it and create a merge request (no prior experience with dart, but I assume the code should not be too hard to find), but since it is such a small change I decided to create the issue in case someone sees it and just fixes it before I find some time to look into a Flutter/Dart application.

And thank you for the awesome app!

### To reproduce

1. Create a PDF file with any tool and rename it to file.PDF 
2. Try to import it in Saber 
3. Gives an error "unsupported filetype"
4. Rename PDF to file.pdf 
5. Import again
6. Works

### Expected behavior

Case of the file extension should not be relevant. Ideally, the filetype would be checked via metadata but I assume this is not easily possible in a cross-platform-app!?

### Saber version

v0.25.6 Google Play (25060)

### Device

- Device: Samsung Galaxy Tab S9
- OS: Native, One UI 7.0, Android 15

### Anything else?

_No response_

ZebraVogel94349:
This fixes #1467, and if related, #1266. Additionally to allowing to import PDF files with uppercase extensions, you can now also import files with uppercase .SBA, .SBN or .SBN2 extensions (or mixed upper-/lowercase extensions).


ZebraVogel94349:
I've just added that all imported sbn and sbn2 files (also those that are in sba files) are saved with lowercase extensions to prevent any problems with uppercase extensions files inside of Saber in the future.

QubaB:
It is important to mention including also #1505 pull request which handles file import on Windows (problem of \ and / in file path).

adil192:
Thanks!

PR #1517, opened 2025-09-04
Update strings_zh_Hans_CN.g.dart

xujiegb:


adil192 (changes requested):
Please only edit `lib/i18n/_missing_translations.yaml` for new translations or `lib/i18n/zh_Hans_CN.i18n.yaml` to replace existing translations.
The `.g.dart` files are generated and any changes therein will be lost.

Issue #1508, opened 2025-08-19
Cant write document into custom directory


Dieter0010:
### Before submitting an issue, please check the following

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

Ive created a directory in the android internal storage root, and assigned saber to that directory.
When i create a new file, add something and try to save it, it simply doesnt save the file.
I have tested this on android 15. 

### To reproduce

1. Create a directory in the android internal storage root.
2. Assign saber to that directory
3. Create a file in saber
4. Click save

### Expected behavior

Save the file

### Saber version

v0.26.2 F-Droid (26020)

### Device

- Device: Galaxy tab S9 FE
- OS: Android 15

zeyuz35:
Seems to be a duplicate/related to #1223

Dieter0010:
No custom dir is working but not if the custom dir is in the android root.

Issue #1411, opened 2024-12-19
Changing folder don't seems to work correctly


Timothee19:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

I wanted to change the folder where my notes ares stored ( to be able to sync through my devices with Syncthing) bu when I did that, the result is just that when I open one of my note on the app, it is note saving anything anymore and I'm "lock" on my note, waiting for the end of saving process, that finally never end, and so I loose my previous notes.
Hopefully I just started to use that app so I loose juste one physics lesson.

### To reproduce

1. Go to "Settings" -> "Advanced" -> "Custom Saber folder" then 
2. Click on the folder icon, choose a new folder (I did in "/storage/data/emulated/0/MyFolder", I can't access to something else than "/storage/data/emulated/0")
3. Then try to open again your note
4. Click on the save button, and it do not work (the saving is not working)

### Expected behavior

I exepected to be able to use my notes as before. And then to use syncthing to synchronise the selected folder with other devices.

### Saber version

v0.25.3 (GooglePlayStore)

### Device

- Device: Samsung Galaxy Tab S9 FE
- OS: Android 14 (One UI 6.1)

### Anything else?

Thanks a lot for everything you do, I hope to be able a day to do same

ZebraVogel94349:
This is a duplicate of #1223 and #1228. To recover your note you could try to downgrade to v0.21.2 using adb as described in #1228. Then export it as an sba file, uninstall Saber, reinstall a newer version and import the file.

jamal-ali1989:
Has this issue been fixed? I also have this issue on Android 15. If i set any other location as a custom folder. I am locked while it says that it is saving data, and I can exit safely once it is over. However, it is just never over.

Issue #1407, opened 2024-11-30
"Selected folder must be empty" after resetting to default


xhalo32:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

I tried to change to a custom saber folder, but it won't change back using the reset button due to an error: "Selected folder must be empty".

### To reproduce

0. I created a folder on the device and export a saber document as `sba` there.
1. I changed "Custom Saber folder" to that folder.
2. I tried to import the `sba` file (after removing the `.bin` suffix) I had just exported which failed for no apparent reason. (Also I'm unable to create new saber documents now.)
3. I restarted the application
4. I tried to revert step 1 by clicking the reset button in the "Custom Saber folder" dialog.
5. The application refused and I lost all my notes :(

### Expected behavior

I expected step 4 (resetting "Custom Saber folder") to work.

I also expect saber to not be in an unusable state (I cannot create new notes or open old ones)

### Saber version

v0.25.2 F-Droid (25020)

### Device

- Device: Samsung Galaxy Tab S8+ (SM-X800)
- OS: Android 14

### Anything else?

I can add a screen recording of the state things are now, but I (probably) can't replicate the bug without losing all my notes. Any help would be appreciated!

xhalo32:
I just noticed the logs, and they say 

```
PathAccessException: Cannot open file, path = '/storage/emulated/0/Android/media.com.github.catfriend1.syncthingandroid/Saber/note.sbn2' (OS Error: Operation not permitted, errno = 1)
```

Maybe using syncthing as a custom saber working directory triggers the bug?

zeyuz35:
I've noticed this annoying bug as well. Custom folders for saber must be completely empty, *including hidden files*, such as those created by syncthing, git, etc.

It is possible to make syncthing, git etc still work, but it is very unwieldy. 
Create a completely empty folder for Saber, and set it as the custom directory, *then*, associate a syncthing/git folder to it. 

FYI saving in custom directories on Android is also bugged in the main branch, see #1223 for discussion and PR that fixes permissions.

ZebraVogel94349:
[...]

Issue #1242, opened 2024-05-03
Files in custom data directory not showing up


bwachtendorf:
### Before submitting an issue, please check the following

- [X] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [X] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

When copying files to the custom data directory, they do not show up in the app, if they were not created there.

### To reproduce

1. Select custom data directory
2. Create note
3. Copy note files from custom data directory somewhere
4. Delete note in app
5. Copy files back to custom data directory

### Expected behavior

All notes that are in the custom data directory should just show up in the app after restart or refresh

### Saber version

v0.21.4 Google Play/F-Droid

### Device

- Device: Samsung S21 FE
- OS: Android 14

### Anything else?

I encountered the issue when trying to open a note that was synced via syncthing.
It is possible to import the note and the app will create a copy (`note (2).sbn2`) and the note appears to be complete.

Iey4iej3:
https://github.com/saber-notes/saber/issues/1223 might be relevant: they might be due to an issue of permission.

TillF:
I observed the same issue.
The solution suggested in #1223 (implemented in pull request from ComputerElite [#1448](https://github.com/saber-notes/saber/pull/1448)) solves the problem for me.
Thanks to all contributors! Please merge these changs to main branch.

PR #1519, opened 2025-09-06
fix: Correct migration procedure

4JX:
Otherwise it keeps creating `oldDir`'s child folders inside themselves until the path is too long for the OS to handle.

*May just have lost about a year's worth of notes to this*. Oh well, lesson learned, backup things more throughoutly.

**Note:** I am really against the delete over at
```dart
await moveDirContents(oldDir: oldDir, newDir: newDir);
await oldDir.delete(recursive: true);
```
Existing because it is a ticking timebomb waiting to delete an user's files if they "Allow access" to the wrong folder. The least that should be done is have a checkbox to *optionally* delete the old stuff with the user's consent. May be able to get worked into this PR, lmk your thoughts.

### DIFF
--- a/lib/data/file_manager/file_manager.dart
+++ b/lib/data/file_manager/file_manager.dart
@@ -16,6 +16,7 @@ import 'package:saber/i18n/strings.g.dart';
 import 'package:saber/pages/editor/editor.dart';
 import 'package:saver_gallery/saver_gallery.dart';
 import 'package:share_plus/share_plus.dart';
+import 'package:path/path.dart' as p;
 
 /// A collection of cross-platform utility functions for working with a virtual file system.
 class FileManager {
@@ -76,7 +77,7 @@ class FileManager {
       log.fine('Old data directory is empty or missing, nothing to migrate');
     } else {
       await moveDirContents(oldDir: oldDir, newDir: newDir);
-      await oldDir.delete();
+      await oldDir.delete(recursive: true);
     }
   }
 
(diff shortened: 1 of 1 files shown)


adil192:
> have a checkbox to _optionally_ delete the old stuff with the user's consent

That's a good idea, probably best to add it in a separate pull request if possible

4JX:
> Thanks for your contribution, especially for tidying up the path handling code.
> 
> I'm still getting this error: `Unhandled Exception: PathAccessException: Cannot copy file to '/storage/emulated/0/sabercustomdir/25-09-23 Untitled.sbn2', path = '/data/user/0/com.adilhanney.saber/app_flutter/Saber/25-09-23 Untitled.sbn2' (OS Error: Operation not permitted, errno = 1)` Would you be able to have a look?

Android is not my strong suit but this is probably in need of a `MANAGE_EXTERNAL_STORAGE` permission check for older android versions that don't have SAF. Some older PRs exist for this, like https://github.com/saber-notes/saber/pull/1448.

What's your device's Android version/API level/model? Did it prompt you to "allow a folder" when you tried to select a custom folder?

For latter tracking, related:
https://github.com/saber-notes/saber/issues/342 (ish)
https://github.com/saber-notes/saber/issues/922
https://github.com/saber-notes/saber/issues/1223
https://github.com/saber-notes/saber/issues/1228
https://github.com/saber-notes/saber/issues/1242
https://github.com/saber-notes/saber/issues/1407
https://github.com/saber-notes/saber/issues/1411
https://github.com/saber-notes/saber/issues/1508

4JX:
> Android is not my strong suit but this is probably in need of a MANAGE_EXTERNAL_STORAGE permission

Ah, seems it is `READ_EXTERNAL_STORAGE`/`WRITE_EXTERNAL_STORAGE` one would want for older android verions. I see `_requestPhotosPermission` has some groundwork for permission checking for `exportFile`, but most of the code elsewhere doesn't appear to check for any kind of permissions before trying.

To add to the other comment's questions is this intentional? Doesn't seem like it'd work to me.

adil192:
> What's your device's Android version/API level/model? Did it prompt you to "allow a folder" when you tried to select a custom folder?

Pixel 8 emulator, Android 16. And yes.
I'll merge this PR for now, and the permissions issue can be handled separately.

[READ_EXTERNAL_STORAGE](https://developer.android.com/reference/android/Manifest.permission#READ_EXTERNAL_STORAGE), [WRITE_EXTERNAL_STORAGE](https://developer.android.com/reference/android/Manifest.permission#WRITE_EXTERNAL_STORAGE), and [MANAGE_EXTERNAL_STORAGE](https://developer.android.com/reference/android/Manifest.permission#MANAGE_EXTERNAL_STORAGE) either aren't allowed or do nothing on newer Android versions. We'd need to use the scoped storage apis

4JX:
> We'd need to use the scoped storage apis

Got my terminology dates mixed up there, yeah meant storage scopes. Initially based my assumption of needing both permission systems off the api level being set to "donut" on the XML, later saw its the build properties file that does it.

> Pixel 8 emulator, Android 16. And yes.

For reference my testing/normal use device is a Samsung tablet on OneUI 6/A14.

Lauloque:
> For latter tracking, related: #342 (ish) #922 #1223 #1228 #1242 #1407 #1411 #1508

Hi, is this PR live in the current release, and does it fix the issues linked above?

It seems risky to test as a user if it effectively results in data locking/loss.

Issue #1524, opened 2025-09-07
Saber can't display any PDF content


carotteatomique:
### Before submitting an issue, please check the following

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Describe the bug

When opening a note originating from an imported PDF, Saber does not display the contents of the PDF, only the notes added by the user.

It also does not display the contents of any newly imported PDF.

From my experience, this affects both v0.26.6 and v0.26.5; v0.26.4 does not seem to be affected and behaves normally.

### To reproduce

1. Open a note originating from an imported PDF
2. Notice the lack of PDF content

Alternatively:
1. Import a new PDF
2. Notice how the pages are all blank

### Expected behavior

It is expected that the PDF content would display.

### Saber version

v0.26.6 (Android, from the GitHub releases through Obtainium)

### Device

- Device: Google Pixel Tablet
- OS: Android 16 (GrapheneOS 2025090300)

### Anything else?

I have also tested on Linux, through the Flatpak. Both the v0.26.6 and v0.26.5 Flatpaks had the same issue, but v0.26.4 was unaffected.

This was tested using a file exported from my tablet, exported using v0.26.6, using the .sba format. This exported file had the same problem on v0.26.6 and v0.26.5 on Linux, but the same file displayed correctly on v0.26.4, leading me to believe this is an issue with Saber and not with the file itself.

The exported PDFs also seem to be affected; a PDF exported from v0.26.6 lacked any of the original PDF's content and looked the same way it displayed in Saber. The PDF exported from v0.26.4, using the same base file, was fine and looked as expected.

This can also be seen in the file sizes:
- Original SBA file: 24,6 Mio
- Correct PDF exported by v0.26.4: 18,0 Mio
- Broken PDF exported by v0.26.6: 2,9 Mio

The extensive tests across versions of Saber and operating systems were only done on one file, but all of my notes originating from PDFs seem to be broken in v0.26.6 on Android.

carotteatomique:
This bug seems to have been fixed in 0.26.7.

Thank you!

PR #1525, opened 2025-09-07
fix: Initialize pdfrx cache directory - important for pdf import

QubaB:
Initialize pdfrx cache directory - so it does not complain it is not set. 

Solves #1524


adil192:
Thank you! Looks like `pdfrxFlutterInitialize` does this as well as some other stuff.

PR #1527, opened 2025-09-07
fix: Correct stow version in pubspec.yaml

QubaB:
Correct stow versions in pubspec.yaml (not commited in 17370e49d9c30320943ff5932376fc45db92de62)

### DIFF
--- a/pubspec.yaml
+++ b/pubspec.yaml
@@ -146,9 +146,9 @@ dependencies:
   pdfrx: ^2.1.3
   path: ^1.9.1
 
-  stow: ^0.5.0
+  stow: ^0.5.1
   stow_plain: ^0.5.0
-  stow_secure: ^0.5.0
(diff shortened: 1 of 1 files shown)


adil192:
Makes no difference, only the pubspec.lock file needs it

QubaB:
I expected pubspec.yaml is origin from which is pubspec.lock updated according to downloaded dependencies. 

When I merged main branch to my Saber version in which I was using different stow versions (my fixed branch) in pubspec.yaml, it was not updated to the newest version.

In fact I am cleaning pubspec.lock on every pull from main branch, because I am using more packages in pubspec.yaml (camera, ...).

adil192:
Ah if you're not using the pubspec.lock from this repo, then it won't carry over. Running `flutter pub upgrade` will handle it, or you could just make this change on your repo

PR #1561, opened 2025-10-18
snap: update to core24 and cleanup

soumyaDghosh:


snap/snapcraft.yaml
```diff
@@ -1,81 +1,61 @@
...
-confinement: strict # use 'strict' once you have the right plugs and slots
+grade: stable
+confinement: strict
 license: GPL-3.0
 compression: lzo
-icon: assets/icon/icon.png
-donation: https://github.com/sponsors/adil192
-source-code: https://github.com/saber-notes/saber
-issues: https://github.com/saber-notes/saber/issues
-contact: soumyadeepghosh2004@zohomail.in
```
adil192: Why were these links removed?

snap/snapcraft.yaml
```diff
@@ -1,81 +1,61 @@
...
-        ln -s ../share/rust/bin/$i $CRAFT_PART_INSTALL/usr/bin/$i
-      done
-    override-prime: |
-      echo 'Skip Prime'
+    plugin: rust
+    source: .
+    rust-channel: "1.90"
+    override-build: ""
+    override-prime: ""
+
```
adil192: Will you need to manually update this rust version?

adil192: The linux github actions workflow doesn't install rust at all. Does the snap build require it?

soumyaDghosh: I am not sure about the dependencies. I just updated the existing deps to get them in a better possible way. I didn't cross check the deps.

snap/snapcraft.yaml
adil192: Nice

adil192:
Thanks @soumyaDghosh

PR #1573, opened 2025-10-27
snap: fix execstack libs and cleanup

soumyaDghosh:
I tested the snap locally and looks like zenity and libjsoncpp is not required any more.


adil192:
Thanks

PR #1576, opened 2025-10-28
i18n: update Simplified Chinese

Jesse205:
I think the result of the automatic translation is smoother than my translation


adil192:
Thanks!

Issue #1569, opened 2025-10-23
Add option to use native window decoration on linux.


Tahosol:
### Before submitting an issue, please check the following

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

### Motivation

<img width="1900" height="258" alt="Image" src="https://github.com/user-attachments/assets/a2c8bc10-f65c-4459-8256-b3b645e5c7bd" />

On some linux setup, the client side decoration look out of place like in my case. The default GTK lib Adwaita window decoration will look out of place for most linux OS except Gnome.

### Solution

I think the best solution would be have an option to use native window decoration on linux, or just enable it as the default behavior. 

### Alternatives

If it is too much to implement a native option, maybe saber can have an option to disable the client side window decoration.

### Saber version

0.26.10(Linux flatpak)

### Device

- Device: Deskctop
- OS: Linux(kernel 6.17.4-arch2-1)

adil192:
I'm not well-versed in Linux window decorations, but as far as I'm aware, it is using the native title bar. You can see in my "before" screenshot, it's using the COSMIC DE titlebar.
To fix the ugliness, I've added a setting (default: turned on) to use a themed (Yaru) titlebar:

| Before | After |
| --- | --- |
| <img width="754" height="1042" alt="Image" src="https://github.com/user-attachments/assets/d7de0933-7757-46d9-9a74-25b949db5cef" /> | <img width="753" height="1040" alt="Image" src="https://github.com/user-attachments/assets/8f2f10eb-79d1-4146-a4a7-ace8f6b6e196" /> |

Tahosol:
Hi @adil192, indeed the decoration right now is Linux‑native, but it is a client‑side decoration, not server‑side. In many desktop environments on Linux, this makes it look out of place, not just ugly. This issue is related to the fact that Flutter uses the GTK 3 header bar, which is technically Linux‑native but not native to desktop environments. On Cosmic, it looks fine because its UI design is highly influenced by GTK; on KDE, it looks somewhat broken. **[flutter/#94381](https://github.com/flutter/flutter/issues/94381)** has an issue on it.

[...]

PR #1578, opened 2025-10-29
tweak: disable CSD on linux

adil192:
This PR disables the client side decorations unconditionally following https://github.com/saber-notes/saber/issues/1569#issuecomment-3457985595.

I've tested COSMIC, Gnome and KDE (on Wayland) and it looks fine with and without the themed title bar setting.
The only side effect I've found is losing the rounded window corners on Gnome and COSMIC, but I think that's fine.

@Tahosol Please can you try running the app from this branch ([link to build instructions](https://github.com/saber-notes/saber/wiki/Build)) and check everything looks right?

Closes https://github.com/saber-notes/saber/issues/1569

### DIFF
--- a/linux/runner/my_application.cc
+++ b/linux/runner/my_application.cc
@@ -1,9 +1,6 @@
 #include "my_application.h"
 
 #include <flutter_linux/flutter_linux.h>
-#ifdef GDK_WINDOWING_X11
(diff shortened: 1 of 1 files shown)


Tahosol:
Thanks, It look great now.
image

PR #1579, opened 2025-10-30

jhihyulin:


adil192:
I'm removing this feature in the next update, so I'll close this PR, but thank you for the translation!

PR #1657, opened 2026-01-31
i18n: Update Czech

ceskyDJ:
Hi @adil192, could you please provide more information about these strings? I less or more guessed the meaning, but I'm still not sure.

### DIFF
--- a/lib/i18n/_missing_translations.yaml
+++ b/lib/i18n/_missing_translations.yaml
@@ -51,10 +51,6 @@ ar:
       watchingServer(OUTDATED): You are currently watching for updates on the server. Editing is disabled in this mode.
       corrupted(OUTDATED): Failed to load note. It may be corrupted or still being downloaded.
 cs:
-  editor:
-    readOnlyBanner:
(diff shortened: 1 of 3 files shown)


adil192:
> Hi @adil192, could you please provide more information about these strings? I less or more guessed the meaning, but I'm still not sure.

Sure. In some situations, editing a note is disabled. Previously, the buttons were just disabled with no explanation why. I've added these strings to tell the user why the note is read only. The full list of reasons can be found in `read_only_reason.dart`:

https://github.com/saber-notes/saber/blob/819252ce110184df1190b36e95bc156155c6758a/packages/sbn/lib/read_only_reason.dart#L1-L25

ceskyDJ:
What a great improvement! Thanks for the details. I think it should be accurate, so feel free to merge it any time.

ceskyDJ:
Btw, are there any variants of these info messages for the rest of the reasons you have in the code?

adil192:
> Btw, are there any variants of these info messages for the rest of the reasons you have in the code?

`placeholder` doesn't show a message at all, since it only occurs in the split second before the editor finishes loading.

The `versionTooNew` banner already existed so there's no new translations needed for that, and `watchingServer` and `corrupted` are covered in this PR.

PR #1678, opened 2026-02-08
i18n: Update Czech

ceskyDJ:


ceskyDJ:
Hi, @adil192! I noticed you modified `README-cs.md` and applied some minor enhancements.

Btw, the last release was kinda messy, as you tagged by PR at some translation from Weblate. Could you please check if the release notes generation works properly?

adil192:
Thanks

> Btw, the last release was kinda messy, as you tagged by PR at some translation from Weblate. Could you please check if the release notes generation works properly?

That was my mistake, I fixed it only a few minutes after though. Is there still an issue?

ceskyDJ:
> That was my mistake, I fixed it only a few minutes after though. Is there still an issue?

Oh, I see. Now, everything is ok.

Issue #1675, opened 2026-02-05
Updated selection tool settings


MilesDawkins:
### Motivation

When using other notes apps, the selection tool has several knobs to turn that are not present in saber notes. Much of my workflow uses the selection tool for smooth operation and i would like to see that in saber. Tools such as Orion Notes and Notewise have these features built in

### Solution

1. Add the ability to have a rectangular selection tool, not just free from
2. Allow the selection tool to select anything that the selection area is touching, not just sections that are fully enclosed. This would allow easy movement of shapes 

### Saber version

v1.29.5 Google Play

### Device

- Device: Galaxy tab S9 FE+
- OS: One UI v8.0, Android v16

### Before submitting an issue, please check the following

- [x] I have searched for similar issues (both open and closed) and cannot find a duplicate
- [x] I agree to follow the [Code of Conduct](https://github.com/saber-notes/saber/blob/main/CODE_OF_CONDUCT.md)

PR #1718, opened 2026-04-06
Add box selection tool

Donut-Muffin:
Previously there was only one lasso selection tool which users would have to drag around object to make selections.

This pull request adds a box selection tool that allows users to make rectangular selections around objects, rather than drawing around them.

Test cases have been implemented (based on the original lasso selection tests). The lasso selection tool has been refactored for clarity.

_Sorry if the project structure needs improvement. This is my first time using Flutter and Dart._

Addresses issue #1675 

https://github.com/user-attachments/assets/de3bb0b3-9475-4ff4-8f54-02b47de89a24


Donut-Muffin:
It appears like a lot of code has been added/removed but a significant portion is refactors since I renamed the Select tool to SelectLasso

PR #1736, opened 2026-04-24
fix: Enable Custom Directories for the FOSS Android build only

aarron-lee:
# Summary 

This PR adds support for Custom Directories for the **FOSS Android build only**

This PR is heavily based on the work done by @ComputerElite in PR #1448 

Note that I'm not an Android developer by trade, I'm primarily a web developer. Please let me know if I overlooked something, or if there's any other problems.

# Implementation details

- Create android-specific Custom Directory Selector
  - create a stub version for the Play store build
  - create a separate FOSS version that implements the external directory prompt
- add a pre script that enables the custom dir for the FOSS build
- update github pipeline to add the required changes for the FOSS build

# Tests performed

Debug builds tested on the following devices:

- S21 Ultra (Android 15)
- Galaxy Fold 3 (Android 15)

Custom dir worked as expected for the FOSS build, also the non-FOSS build behaved as-expected.

## Partially fixes:
Issue #1223

### DIFF
--- a/.github/workflows/android.yml
+++ b/.github/workflows/android.yml
@@ -236,6 +236,9 @@ jobs:
       - run: flutter pub get
       - run: ./patches/post/patch_rust_versions.sh
       - run: ./patches/post/remove_wasm_libs.sh
+      - name: Apply custom-dir changes to FOSS build
+        if: matrix.foss
+        run: ./patches/pre/enable_custom_dir_foss_android.sh
 
       - name: Build apk
         run: flutter build apk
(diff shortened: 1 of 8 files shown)


adil192:
This is just a hack, not an actual fix.
Giving an app (even a FOSS one) complete access to all your files is a bad idea.
If you wish to fix this without a hack, you should debug and fix our scoped storage implementation.

aarron-lee:
Makes sense.

Unfortunately I don't have any experience with Android development, so hopefully someone else with more knowledge could chime in for scoped storage then.

Thank you for taking a look at the PR.

PR #1743, opened 2026-05-07
snap: add llvm and clang to fix build

soumyaDghosh:


adil192:
thanks! the build succeeds again

PR #1763, opened 2026-06-06
feat: YaruWindowTitleBar implement

airon-main:
Before:
image

After:
image

---
## Summary by cubic
Added a custom Windows title bar with `YaruWindowTitleBar`, showing the Saber SVG icon and app title. Windows-only initialization and UI wrapper; macOS/Linux/other platforms are unchanged.

- **Refactors**
  - Run `YaruWindowTitleBar.ensureInitialized()` at startup in `main.dart` only on Windows.
  - Use `MaterialApp.builder` in `ExplicitlyThemedApp` to wrap with `_WindowsTitleBarWrapper` on Windows, composing with any existing `builder`.

Note: Saber is trialing Cubic AI code review. AI output may contain mistakes, misconceptions, and hallucinations. You can act on the AI feedback if you find it helpful and are sure it's correct; otherwise you can disregard it and wait for a human reviewer.

<sup>Written for commit daf93b6689e51ff2dc15cbd236bfe64ea7afff16. Summary will update on new commits.</sup>


airon-main:
YO YARU IS GEM, HOW DIDN'T I KNOW THIS

airon-main:
ishowspeed-talking-tom-news

lib/main.dart
```diff
@@ -46,6 +47,8 @@ Future<void> main(List<String> args) async {
 
 Future<void> appRunner(List<String> args) async {
   WidgetsFlutterBinding.ensureInitialized();
+  if (Platform.isWindows || Platform.isLinux || Platform.isMacOS)
+    await YaruWindowTitleBar.ensureInitialized();
```
adil192: This should be moved into the `await Future.wait([` block on line ~90 so it can be awaited in parallel to avoid slowing down app startup.
Also it should only run on Windows, not Linux/macOS.

lib/components/theming/dynamic_material_app.dart
```diff
@@ -91,6 +93,31 @@ class DynamicMaterialAppState extends State<DynamicMaterialApp>
...
+              ],
+            ),
+            centerTitle: false,
+            buttonPadding: const EdgeInsets.only(bottom: 1),
+          ),
+          body: child,
+        );
+      }
+      return builder;
+    }
```
adil192: This is a little inelegant, especially because we're passing the same arguments to `ExplicitlyThemedApp` each time. We should move this stuff into `ExplicitlyThemedApp` instead of `DynamicMaterialApp` to avoid repetition.

I'd recommend moving this widget tree into its own StatelessWidget for performance reasons. (Flutter can skip rebuilding a widget when its arguments remain the same.)
```dart
class _WindowsTitleBarWrapper extends StatelessWidget {
  const _WindowsTitleBarWrapper({required this.child});

  final Widget child;

  @override
  Widget build(BuildContext context) {
    return Scaffold(.....);
  }
}
```

And then in `ExplicitlyThemedApp`, pass `builder` like this:
```dart
  return MaterialApp.router(
    // ...
    builder: theme.platform == .windows
        ? (context, child) => _WindowsTitleBarWrapper(child: child)
        : null,
  );
```

adil192 (changes requested):
Your screenshots look nice, thanks. Just some nitpicks to address now

airon-main:
Was procastinating :(

airon-main:
ishowspeed-talking-tom-news

